PR #5, opened 2021-07-12
Add long config flag to Viceroy

### LINKED ISSUE #3: Introduce a long `--config` alias for `-C`
Currently, we provide a `-C` option for providing the path to configuration for Viceroy (_see below_). We should add a longer `--config` option for those who prefer that style.

https://github.com/fastly/Viceroy/blob/main/cli/src/opts.rs#L33

Reported by @peterbourgon. :slightly_smiling_face:

mgattozzi:
We only had the short `-C` flag to pass in a config for Viceroy, but
some people would prefer a longer `--config`. This is also useful for
those learning how to use Viceroy to know what the flags mean, rather
than their shorthand version.

Closes #3

### DIFF
--- a/cli/src/opts.rs
+++ b/cli/src/opts.rs
@@ -30,7 +30,7 @@ pub struct Opts {
     )]
     input: PathBuf,
     /// The path to a TOML file containing `local_server` configuration.
-    #[structopt(short = "C")]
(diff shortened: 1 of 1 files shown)


cratelyn (approved):
Thanks!

PR #27, opened 2021-07-12
📦 Align release artifacts with the CLI

aturon:
- Produce .tar.gz archives
- Include version number in archive name
- Remove Windows for the time being

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,42 +21,22 @@ jobs:
           default: true
           override: true
 
-      - name: Build
-        run: cargo build --all --release && strip target/release/viceroy && mv target/release/viceroy target/release/viceroy_linux-amd64
+      - name: Extract tag name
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -76,13 +56,21 @@ jobs:
...
 
-      - name: Build
-        run: cargo build --all --release && strip target/release/viceroy && mv target/release/viceroy target/release/viceroy_darwin-amd64
+      - name: Extract tag name
+        uses: olegtarasov/get-tag@v2.1
+        id: tagName
+
+      - name: Build and package
+        run: |
+          cargo build --all --release
```
cratelyn, line 65: Could we also use `--locked` here, so that we never create release binaries that wouldn't have an up-to-date lockfile?

aturon: Added!

cratelyn (approved):

Issue #28, opened 2021-07-12
[Release] Prepare for 0.2.1 release


cratelyn:
**EDIT**: this was supposed to be a PR, not an issue! How embarrassing.

PR #30, opened 2021-07-12
🐦 update crates to new development version (0.2.2)

cratelyn:
Follow-up to #28, bumps our in-tree versions so that they reflect what the next release will be.


aturon (approved):
🎸

PR #32, opened 2021-07-13
Replace mandatory extension with validation

### LINKED ISSUE #2: Remove restriction on `.wasm` extension
Currently, Viceroy expects that the given binary has a `.wasm` extension. This causes hassle for some users, aside from the fact that `.wat` text format programs would also be acceptable to Viceroy. Let's remove this restriction, and instead provide a helpful error message in the event of an invalid Wasm binary.

Reported by @fgsch.

fgsch:
Rather than restricting the extension to .wasm and later failing when loading the file, accept any file but validate its content.

As a bonus, disambiguate OS specific errors (e.g., file does not exist) from invalid format.

Fixes #2.


aturon (approved):
This is awesome, thank you! 💯

PR #33, opened 2021-07-13
Allow more functions per wasm module

tschneidereit:
Some services can contain a very large number of functions, in particular when compiled in debug mode, so to debug them in Viceroy requires a higher limit


cratelyn (approved):
Thank you!

PR #34, opened 2021-07-13
Workaround bug in clap

fgsch:
If the default_value is specified, clap shows --addr as a positional
argument in the usage output and not as part of the options.


fgsch:
Before:

```
error: The following required arguments were not provided:
    <input>

USAGE:
    viceroy <input> --addr <socket-addr>

For more information try --help
```

After:

```
error: The following required arguments were not provided:
    <input>

USAGE:
    viceroy [FLAGS] [OPTIONS] <input>

For more information try --help
```

cli/src/opts.rs
```diff
@@ -51,6 +51,7 @@ impl Opts {
     /// The address that the service should be bound to.
     pub fn addr(&self) -> SocketAddr {
         self.socket_addr
+            .unwrap_or_else(|| "127.0.0.1:7878".parse().unwrap())
```
cratelyn, line 54: nit: Would you mind constructing the `SocketAddr` via [`new`](https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html#method.new), rather than a `.parse().unwrap()` chain?

cratelyn (changes requested):
I have just a _teeny_ tiny nitpick about how we provide our default value. Thank you @fgsch for this thoughtful change, I'm glad that we will have better usage output now :tada:

cli/src/opts.rs
fgsch: I originally had that but decided to use the parse method to keep the textual representation so it's easier to search for people unfamiliar with rust but I can revert back to Sockaddr::new()..

fgsch: Updated.

cratelyn (dismissed):
Thank you!!

aturon (approved):

PR #35, opened 2021-07-14
Make GH Actions also run on MacOS and Windows

### LINKED ISSUE #6: Add required CI checks for Mac, Windows
Currently, our CI system is a little bit minimal! We only have a job that runs on Ubuntu, with CI coverage lacking for other platforms that we hope to support. Let's flesh out the matrix of platforms we run checks on, here: https://github.com/fastly/Viceroy/blob/main/.github/workflows/test.yml#L8

mgattozzi:
We support all Tier 1 platforms that Rust uses, but we have not been
testing that they also build on other platforms besides Linux. With this
change CI will also make sure that a build will work on these platforms
as well, letting us catch any OS specific breakages that we might not
catch developing Viceroy on our own machines.

Closes #6

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       matrix:
         rust-toolchain: [1.52.1]
-        platform: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


aturon (approved):
Thank you!

PR #37, opened 2021-07-14
🧹 Tidy up logging and error output

aturon:
This PR addresses a few minor UI nits (reported on Slack), as well as a more significant issue: rather than panicking when an unimplemented hostcall is used, we instead yield an error back to the guest. The error reporting infrastructure will print a friendly message, though in most cases the SDK will subsequently panic.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -19,7 +19,7 @@ use {
         sync::Arc,
     },
     tokio::sync::oneshot::{self, Sender},
-    tracing::{event, info_span, Instrument, Level},
+    tracing::{event, info, info_span, Instrument, Level},
     wasmtime::{Engine, InstancePre, Linker, Module},
 };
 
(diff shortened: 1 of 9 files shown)


lib/src/execute.rs
```diff
@@ -281,9 +276,8 @@ impl ExecuteCtx {
             .expect("`memory` is exported")
             .size(&store);
 
-        event!(
-            Level::INFO,
-            "Request completed, using {} bytes of wasm heap",
+        info!(
+            "request completed using {} bytes of WebAssembly heap",
```
phamann, line 280: Maybe a future improvement and out of scope of this PR, however, it would be nice if the `n bytes` could be a more human-friendly representation I.e. 1 Kb, 1Mb, 1Gb  etc. Such as [bytesize](https://docs.rs/bytesize/1.0.1/bytesize/).

aturon: Sounds great, I'll try to sneak that one in as well!

aturon: Alright, I added a commit to make the heap size human-friendly, thank you!

mgattozzi (approved):
These changes look good to me. I'm glad we have a better error now to let people know we don't support something on local testing yet!

cratelyn (approved):

PR #38, opened 2021-07-15
Have CI use --locked flag for cargo commands

### LINKED ISSUE #15: CI should enforce lockfiles
We should be sure that CI enforces our lockfiles when building release binaries, and when checking pull requests.

See https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries for more information.

mgattozzi:
We were already making sure that our releases checked that Viceroy was
using an updated Cargo.lock file, but it would be better if we checked
this on every PR as well so that we can release code without any
hiccups. This commit changes our Makefile which is used as part of CI to
use the --locked flag for the cargo commands that could use it so that
each PR is checked for an up to date lockfile.

Closes #15


aturon (dismissed):

mgattozzi:
@aturon ♻️ ✅ as I rebased off the CI fixes!

aturon (approved):

cratelyn:
Well, we know that it works! :rofl: 

```
error: the lock file /home/runner/work/Viceroy/Viceroy/cli/tests/trap-test/Cargo.lock needs to be updated but --locked was passed to prevent this
```

aturon (approved):

PR #41, opened 2021-07-15
🚢 update viceroy package name and README

### LINKED ISSUE #31: README mentions `--manifest-path` but that flag doesn't appear in the help output

aturon:
This PR addresses a couple remaining issues for release and publication to crates.io: updating the cli package name to `viceroy`, and updating the README to reflect the current state of play.

Fixes #31.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,27 @@ Rust library that can be embedded into your own testing infrastructure.
 
 ## Installation
 
-To install Viceroy, you'll need to first
+### Via the Fastly CLI
+
+As mentioned above, most users of Compute@Edge should do local testing via the
+Fastly CLI, rather than working with Viceroy directly. Any [CLI release] of
+version 0.34 or above supports local testing, and the workflow is documented
+[here][cli].
+
+[CLI release]: https://github.com/fastly/cli/releases
+
+### As a standalone tool from crates.io
+
+To install Viceroy as a standalone tool, you'll need to first
 [install Rust](https://www.rust-lang.org/tools/install) if you haven't already.
 Then run `cargo install viceroy`, which will download and build the latest
 Viceroy release.
 
-## Usage
+## Usage as a standalone tool
+
+**NOTE**: the Viceroy standalone CLI has a somewhat different interface from that
+of [the Fastly CLI][cli]. Command-line options below describe the standalone
+Viceroy interface.
 
 After installation, the `viceroy` command should be available on your path. The
 only required argument is the path to a compiled `.wasm` blob, which can be
--- a/README.md
+++ b/README.md
@@ -5,48 +5,19 @@ allows you to run services written against the Compute@Edge APIs on your local
 development machine, and allows you to configure testing backends for your
 service to communicate with.
 
-In the near future, Viceroy will be packaged as part of the Fastly CLI and
-integrate with other Compute@Edge workflows. But for the moment, you can use it
-by hand, as described below.
+Viceroy is normally used through the [Fastly CLI's `fastly compute serve`
+command][cli], where it it fully integrated into the Compute@Edge workflows.
+However, it is also a standalone open source tool with its own CLI and a
+Rust library that can be embedded into your own testing infrastructure.
+
+[cli]: https://developer.fastly.com/learning/compute/testing/#running-a-local-testing-server
 
 ## Installation
 
 To install Viceroy, you'll need to first
 [install Rust](https://www.rust-lang.org/tools/install) if you haven't already.
-
-### Clone the repo
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -5,48 +5,19 @@ allows you to run services written against the Compute@Edge APIs on your local
 development machine, and allows you to configure testing backends for your
 service to communicate with.
 
-In the near future, Viceroy will be packaged as part of the Fastly CLI and
-integrate with other Compute@Edge workflows. But for the moment, you can use it
-by hand, as described below.
+Viceroy is normally used through the [Fastly CLI's `fastly compute serve`
+command][cli], where it it fully integrated into the Compute@Edge workflows.
```
cratelyn, line 9:
```suggestion
command][cli], where it is fully integrated into Compute@Edge workflows.
```

README.md
```diff
@@ -5,48 +5,19 @@ allows you to run services written against the Compute@Edge APIs on your local
...
-### Updating Viceroy
-
-To update Viceroy, navigate to the `Viceroy` directory you cloned above, and then:
-
-```
-git pull
-git submodule update --recursive --init
-cargo install --path cli
-```
+Then run `cargo install viceroy`, which will download and build the latest
```
cratelyn, line 19: First, to state the obvious: This means we'll need to publish our crates before we merge this.

README.md
```diff
@@ -5,48 +5,19 @@ allows you to run services written against the Compute@Edge APIs on your local
...
+command][cli], where it it fully integrated into the Compute@Edge workflows.
+However, it is also a standalone open source tool with its own CLI and a
+Rust library that can be embedded into your own testing infrastructure.
+
+[cli]: https://developer.fastly.com/learning/compute/testing/#running-a-local-testing-server
 
 ## Installation
 
 To install Viceroy, you'll need to first
 [install Rust](https://www.rust-lang.org/tools/install) if you haven't already.
```
cratelyn, line 18: :grey_question: What would you think about breaking this into two sections: (a) "_building from source_", which would be what's already here, and (b) "_releases_"  mostly just pointing to our [release page](https://github.com/fastly/Viceroy/releases)?

It might also be worth mentioning in this section that the Fastly CLI will install this for you automatically. These instructions are just for people hoping to use it independently of the CLI, e.g. within testing infrastructure.

cratelyn (commented):
Huzzah! I have one (1) typo to fix, and then a small quibble about the "_Installation_" section.

Otherwise, this looks like it's ready to go!

README.md
aturon: Or just after :-)

Perhaps the package rename should be broken out into a separate PR if we want to publish first.

README.md
aturon: Hmm I have mixed feelings about keeping around the old docs re: installation.

I think part of what's giving me pause is: I don't really know who our audience is, here. We generally don't want customers to use Viceroy directly. OTOH folks hacking directly on it probably already know relevant Rust workflows around installation etc.

Do you have thoughts on audience/goals here?

README.md
cratelyn: D'oh! Right, because the rename is in this diff. :sweat_smile: 

I'll mark this as resolved, and we can handle publishing after this lands.

README.md
cratelyn: > **[..]** keeping around the old docs **[..]**

Oh my goodness pardon me! To clarify, when I said "_what's already here_" I meant the state of our readme _after_ these changes.

As for audience, I agree that we don't expect (_or want?_) people to use Viceroy directly, without good reason. So I'd hope that this section redirects the general majority of people that will (most likely) need to upgrade their CLI rather than install this tool directly.

The people sticking around after that disclaimer we might assume have some specific use-case that warrants using Viceroy directly, and can then choose from either a release binary or building/installing from source via `cargo install`.

How's that sound?

cratelyn: Building on this: maybe we want to name-drop that people will want a CLI version >= `0.34.0`, and then point them [here](https://github.com/fastly/cli/releases) if they need to upgrade.

aturon: Ahhh sorry for the misunderstanding! Yes, this makes a ton of sense, I will update accordingly.

cratelyn: No apology needed, thank you for asking the question and forcing me to hash this out a bit more explicitly! We're better off for it. :blush:

aturon:
I've pushed another commit, refining the text to push people more toward the Fastly CLI, and to be more clear that the CLI descriptions in the README are about Viceroy's standalone interface.

cratelyn (approved):
Thank you! Now, off to the registry :heart_eyes: :package: :bookmark_tabs:

PR #44, opened 2021-07-15
Use wasmtime-wasi in async mode

pchickey:
Previously, we linked with the synchronous version of the Wasi import functions, but used the tokio implementations inside them. This led to crashes in wiggle's "dummy executor", which shouldn't be in the picture.

Now we link import functions with `wasmtime_wasi::tokio`'s implementation, which defers to the underlying async engine. And, a test is added to show that this fixes the issue.

### DIFF
--- a/cli/tests/sleep.rs
+++ b/cli/tests/sleep.rs
@@ -18,5 +18,3 @@ async fn empty_ok_response_by_default_after_sleep() -> TestResult {
 
     Ok(())
 }
-
-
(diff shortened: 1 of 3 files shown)


aturon (approved):

cratelyn:
@pchickey looks like we'll need a quick `cargo fmt` pass, apologies!

aturon (approved):

Issue #9, opened 2021-07-12
Allow host overrides in TOML configuration


cratelyn:
One option exposed in Fastly's management UI is the ability to provide an "_override host._" See [this guide](https://docs.fastly.com/en/guides/specifying-an-override-host) for more information. The core premise is that the `host` header will be rewritten with a _different_ host than is in the outbound URL.

We should expose a setting in our TOML schema to do the same thing when running a program locally.

aturon:
@fgsch plans to take this on!

Some notes:
* Backend configuration is defined [here](https://github.com/fastly/Viceroy/blob/813b076da9e1b5a7ed1f615ee80ffeeaac8c2cff/lib/src/config/backends.rs#L9-L12), and deserialized [here](https://github.com/fastly/Viceroy/blob/813b076da9e1b5a7ed1f615ee80ffeeaac8c2cff/lib/src/config/backends.rs#L82-L94).
  * Feel free to choose a new key to sit beside `url` for the optional override host.
* We already do [some surgery on URLs when making backend requests](https://github.com/fastly/Viceroy/blob/813b076da9e1b5a7ed1f615ee80ffeeaac8c2cff/lib/src/upstream.rs#L60-L87). That's probably where this host override logic should go.

cratelyn:
I think this issue is done. Thanks again Federico!

PR #48, opened 2021-07-19
Allow host overrides in TOML configuration

fgsch:
This adds a new field to the backend definition, override_host,
allowing to override the Host header that will be sent to the
backend.

Addresses #9.

### DIFF
--- a/cli/tests/upstream.rs
+++ b/cli/tests/upstream.rs
@@ -2,7 +2,10 @@ mod common;
 
 use {
     common::{Test, TestResult},
-    hyper::{header, Request, Response, StatusCode},
+    hyper::{
+        header::{self, HeaderValue},
+        Request, Response, StatusCode,
+    },
 };
 
 #[tokio::test(flavor = "multi_thread")]
@@ -129,7 +132,7 @@ async fn override_host_works() -> TestResult {
         .host(9000, |req| {
             assert_eq!(
                 req.headers().get(header::HOST),
-                Some(&hyper::header::HeaderValue::from_static("otherhost.com"))
+                Some(&HeaderValue::from_static("otherhost.com"))
             );
             Response::new(vec![])
         });
--- a/cli/tests/upstream.rs
+++ b/cli/tests/upstream.rs
@@ -2,7 +2,7 @@ mod common;
 
 use {
     common::{Test, TestResult},
-    hyper::{Request, Response, StatusCode},
+    hyper::{header, Request, Response, StatusCode},
 };
 
 #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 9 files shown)


cli/tests/upstream.rs
```diff
@@ -13,18 +13,31 @@ async fn upstream_sync() -> TestResult {
...
+        // The "override-host" backend checks the Host header
+        .backend(
+            "override-host",
+            "http://127.0.0.1:9002/",
+            Some("otherhost.com"),
+        )
+        .host(9002, |req| {
+            assert_eq!(
+                req.headers().get(header::HOST),
+                Some(&hyper::header::HeaderValue::from_static("otherhost.com"))
```
cratelyn, line 38: Sorry, import nit: Could we remove the fully-specified identifier, and just use `HeaderValue` here? It's not ambiguous with other symbols here.

cratelyn:
```suggestion
                Some(&HeaderValue::from_static("otherhost.com"))
```

cli/tests/upstream.rs
```diff
@@ -2,7 +2,7 @@ mod common;
 
 use {
     common::{Test, TestResult},
-    hyper::{Request, Response, StatusCode},
+    hyper::{header, Request, Response, StatusCode},
```
cratelyn, line 5:
```suggestion
    hyper::{header::HeaderValue, Request, Response, StatusCode},
```

See import comment below :point_down: (_Unsure if this will cause some `cargo fmt` noise_)

lib/src/config/unit_tests.rs
```diff
@@ -183,15 +184,63 @@ mod testing_config_tests {
...
+            res => panic!("unexpected result: {:?}", res),
+        }
+    }
+    /// Check that override_host field is valid.
+    #[test]
+    fn backend_configs_must_provide_a_valid_override_host() {
+        use BackendConfigError::InvalidOverrideHost;
+        static BAD_OVERRIDE_HOST_FIELD: &str = r#"
+            [backends]
+            "shark" = { url = "http://a.com", override_host = "somehost.com\n" }
```
cratelyn, line 236: A deep thank you for filling out thorough unit tests here :pray:

cli/tests/upstream.rs
```diff
@@ -100,6 +113,20 @@ async fn upstream_sync() -> TestResult {
         "/hello/greeting.html"
     );
 
+    ////////////////////////////////////////////////////////////////////////////////////
+    // Test that override_host works as intended
+    ////////////////////////////////////////////////////////////////////////////////////
+
+    let resp = test
+        .against(
```
cratelyn, line 121:
```suggestion
    let resp = test
        .via_hyper()
        .against(
```

Can we use `Test::via_hyper` here, so we're also sending this request through Hyper? It ought not to affect the result, but it'd bring some additional peace of mind that we won't see any regressions :slightly_smiling_face:

cratelyn (commented):
:raised_hands: :partying_face: Woohoo! Thanks for taking this issue on @fgsch, this all looks wonderful! I have two _very_ small nitpicks to address, one concerning a `use` import, and a one-line alteration to our new integration test.

Hopefully, this should all be addressable from the web UI, and we can get this landed soon!

cli/tests/upstream.rs
fgsch: Do you mean import HeaderValue directly? btw, http-semantics.rs uses the full-specified identifier.

cli/tests/upstream.rs
fgsch: I need header as well for HOST, btw.

cratelyn: `header::{self, HeaderValue}` would work, in that case.

fgsch: Yup, I know. I will change it but isn't this subjective?

aturon: It is subjective but FWIW there's a pretty strong Rust idiom to prefer imports to qualified paths in the vast majority of cases (unless some kind of disambiguation is needed).

fgsch:
I've rebased this PR to fix the via_hyper() commit and drop the commit that was cherry-picked in #49.
Sorry, will be more careful next time to avoid rebasing it after it is been reviewed.

cli/tests/upstream.rs
fgsch: Updated.

cratelyn (approved):
Wahoo! Thanks for implementing this @fgsch :slightly_smiling_face:

PR #51, opened 2021-07-20
🧹 Eliminate duplicate documentation

aturon:
The README was duplicating some of the contents of the official CLI docs. This PR replaces that material with pointers to the devhub docs.


cratelyn (approved):
Thanks!!

PR #60, opened 2021-08-06
add initial work for dictionary support

JakeChampion:
- Done: Reading dictionaries from fasty.toml file
- Done: Construct Dictionary and DictionaryHandles
- Done: Implement Dictionary.open
- To do: Reading the dictionary items from fastly.toml file
- To do: Implement Dictionary.get

Feedback for Fastly so far:
It would be really useful to have a few more methods on Dictionary such as:
- `has` to check the existence of an key
- `forEach` to iterate over the items
- `getStream` to read the item value as a stream, this would be good to keep memory use low

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2112,6 +2112,7 @@ dependencies = [
  "hyper-tls",
  "itertools",
  "lazy_static",
+ "once_cell",
  "regex",
  "semver 0.10.0",
  "serde",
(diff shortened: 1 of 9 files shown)


JakeChampion:
Apologies, I created this pull-request on the wrong repository.
I forgot GitHub opens the pull-request on the upstream instead of on the fork.

Issue #11, opened 2021-07-12
Implement edge dictionary support


cratelyn:
As mentioned in the project's [README](https://github.com/fastly/Viceroy/blob/main/README.md#limitations), we do not currently support all of the features provided by Compute@Edge's production environment. One of these missing features is Edge Dictionaries.

This is a tracking issue to document ongoing work to implement this support.

JakeChampion:
The limitations section in the readme was removed in [commit ba3d8a4](https://github.com/fastly/Viceroy/commit/ba3d8a46a22670d89bf8b2f5dff22d266800d2cd#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

This is what it said:

> ## Limitations
> 
> At the moment, Viceroy does not support the full Compute@Edge API. In
> particular:
> 
> * GeoIP is unsupported.
> * Dictionaries are unsupported.
> * Caching directives are ignored; no caching is ever performed.
> * Information about the TLS connection from the client is not available.
>

PR #61, opened 2021-08-09
Add local dictionary support

JakeChampion:
Resolves https://github.com/fastly/Viceroy/issues/11

This reads a new section from the `local_server` part of fastly.toml named `dictionaries`.

The `local_server.dictionaries` section is an array of tables with two fields: name, and file.

The name field is a string which is used as the name of the dictionary and is used via the Dictionary.open method. This field uses the same validation rules asthe Fastly API (Name must start with alphabetical and contain only alphanumeric, underscore, and whitespace).

The file field is a string which is a file-path pointing to a JSON file which contains all the dictionary items within an object.

An example of a full local_server section of fastly.toml now looks like this:
```toml
[local_server]
  [local_server.backends]
    [local_server.backends.dog]
      url = "http://localhost:7878/dog-mocks"
  [local_server.dictionaries]
    [local_server.dictionaries.a]
      file="./a.json"
      format = "json"
```

---

**Note for reviewers:** 

I couldn't figure out how the tests are implements for the wiggle abi parts such as `lib/src/wiggle_abi/dictionary_impl.rs`. I assume they are implemented in an internal/private repo that Fastly has?

---

- [X] Reading dictionaries from fasty.toml file
	- Note: I added the local dictionaries to the `local_server` section
- [x] Construct Dictionary and DictionaryHandles
	- [x] Dictionaries have two fields: name, and file
	- [x] Validate name field with Fastly' validation rules (Name must start with alphabetical and contain only alphanumeric, underscore, and whitespace)
	- [x] Validate file file (perhaps by doing a file existence check and checking it is JSON which adheres to our schema)
- [x] Implement Dictionary.open
- [x] Implement Dictionary.get
- [x] Tests
- [ ] documentation


lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -9,20 +12,63 @@ use {
     wiggle::GuestPtr,
 };
 
+use memoize::memoize;
+
+#[memoize]
```
JakeChampion, line 17: I added `memoize` here to avoid reading the file multiple times.

I think a better solution may be to have the file get read into memory when the `FastlyDictionary.open` method is called instead of reading it when the `FastlyDictionary.get` method is called (which we currently do).

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -9,20 +12,63 @@ use {
...
+fn read_json_file(file: String) -> serde_json::Map<String, serde_json::Value> {
+    let data = fs::read_to_string(file).expect("Unable to read file");
+    let json: serde_json::Value = serde_json::from_str(&data).expect("JSON was not well-formatted");
```
JakeChampion, line 20: We would need to document somewhere what the expected schema is for the JSON file. Perhaps on [the docs site](https://developer.fastly.com/reference/fastly-toml/#local-server)?

The schema currently is `Object<String, String>`

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -9,20 +12,63 @@ use {
...
+                let item = serde_json::to_string(item).unwrap();
+                let item_bytes = item.as_bytes();
+
+                if item_bytes.len() > buf_len as usize {
+                    // Write out the number of bytes necessary to fit this dictionary_item, or zero on overflow to
+                    // signal an error condition.
+                    buf.write(item_bytes.len().try_into().unwrap_or(0))?;
+                    return Err(Error::BufferLengthError {
+                        buf: "dictionary_item",
+                        len: "dictionary_item_max_len",
```
JakeChampion, line 61: It's not possible via the Fastly API to create a dictionary item that is larger than `dictionary_item_max_len`, which means this is only a possible runtime error when using Viceroy.

Should we have Viceroy check all the dictionary items are below `dictionary_item_max_len` prior to starting up the server? It would change this from being a runtime error to a compile time error, giving users a shorter feedback loop

lib/src/wiggle_abi/dictionary_impl.rs
peterbourgon: > I added `memoize` here to avoid reading the file multiple times.

I reckon it would be nice if you could update the dictionary and see the changes without restarting Viceroy?

> I think a better solution may be to have the file get read into memory when the FastlyDictionary.open method is called instead of reading it when the FastlyDictionary.get method is called (which we currently do).

In real Compute@Edge, the dictionary is lazy-loaded on Get, more or less, so the existing behavior here seems most symmetric.

lib/src/wiggle_abi/dictionary_impl.rs
peterbourgon: If you want to jam some arbitrary bytes into a JSON string, typically you'd base64-encode them. Will serde_json handle that properly, i.e. give you the actual bytes back rather than their base64 encoding?

lib/src/wiggle_abi/dictionary_impl.rs
JakeChampion: I've added this in [`a61a697` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/a61a6978eae28d216dc014a30ce982c6cc62e7eb) 👍

cratelyn (changes requested):
:heart_eyes_cat: This is incredible stuff! Thank you for opening this PR.

I have left assorted comments below; the two high-level points are that we should think about how we deserialize the configuration, and how we assign handles.

lib/src/config/unit_tests.rs
```diff
@@ -243,4 +282,118 @@ mod testing_config_tests {
...
+        }
+    }
+
+    /// Check that dictionary definitions cannot contain unrecognized keys.
+    #[test]
+    fn dictionary_configs_cannot_contain_unrecognized_keys() {
+        use DictionaryConfigError::UnrecognizedKey;
+        static BAD_DEFAULT: &str = r#"
+            [dictionaries]
+            thing = { name = "thing", file = "./file.json", shrimp = true }
```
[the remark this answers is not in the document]
iximeow: ya love to see a real shrimp-:a:ck in the wild

lib/src/wiggle_abi/dictionary_impl.rs
peterbourgon: Ah, sorry! 😅 I'm not asking for a change or anything, I'm just asking about the present behavior. But I guess Edge Dictionaries currently model values as strings rather than arbitrary bytes, so there's no reason to do anything differently here. Sorry for the distraction!

cli/src/main.rs
```diff
@@ -13,14 +13,16 @@
 #![cfg_attr(not(debug_assertions), doc(test(attr(allow(dead_code)))))]
 #![cfg_attr(not(debug_assertions), doc(test(attr(allow(unused_variables)))))]
 
+use std::convert::TryInto;
+
```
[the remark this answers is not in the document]
JakeChampion: You're right, Rust Analyzer adding the imports automatically ☺️

cli/src/main.rs
```diff
@@ -88,6 +93,52 @@ pub async fn serve(opts: Opts) -> Result<(), Error> {
                 Ok(_) => event!(Level::INFO, "backend '{}' is up", name),
             }
         }
+
+        let dictionary_max_len: usize = 1000;
+        let dictionary_item_key_max_len: usize = 256;
+        let dictionary_item_value_max_len: usize = 8000;
+        for (name, dictionary) in dictionaries.iter() {
```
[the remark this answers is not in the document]
JakeChampion: Good point about this code not being in `main.rs`, I moved the error handling into `viceroy_lib::config::dictionaries::Dictionary`, specifically I moved it into `impl TryFrom<Table> for Dictionary` -- this was done in [`541650a` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/541650aaad0eea96c317976708aaa401f35d0f3b)

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,143 @@
...
+/// A single Dictionary definition.
+///
+/// A Dictionary consists of a name and an id, but more fields may be added in the future.
+#[derive(Clone, Debug)]
+pub struct Dictionary {
+    // Name must start with alphabetical and contain only alphanumeric, underscore, and whitespace
+    pub name: DictionaryName,
+    pub id: DictionaryHandle,
+    pub file: String,
+}
```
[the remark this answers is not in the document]
JakeChampion: >Separately: I think we can remove the name field here, and consequentially in the TOML. The way that these are keyed in [local_server.dictionaries.name] should give us the same information provided we set up our #[serde::Deserialize] types appropriately.

I really wanted to achieve that before making the pull-request but I can't figure out how to implement this functionality 😔

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// A map of [`Backend`] definitions, keyed by their name.
+#[derive(Clone, Debug, Default)]
+pub struct DictionariesConfig(pub HashMap<String, Arc<Dictionary>>);
```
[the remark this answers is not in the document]
JakeChampion: Good idea, I've done this now 👍

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,143 @@
...
+    std::{
+        collections::HashMap,
+        sync::{atomic::AtomicU32, Arc},
+    },
+};
+
+static COUNTER: AtomicU32 = AtomicU32::new(1);
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DictionaryName(pub String);
```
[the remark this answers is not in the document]
JakeChampion: I've removed the internal pub and implemented `Display` and `Hash` -- `Display` is required because we of the errors and tracing in [`lib/src/config/dictionaries.rs`](https://github.com/fastly/Viceroy/pull/61/files#diff-07a2e6099d701192a7ccd4206311fc29081f2b6d753b4427286f450b3b874b3eR133-R137)

lib/src/session.rs
```diff
@@ -67,6 +69,14 @@ pub struct Session {
...
     /// Populated prior to guest execution, and never modified.
     pub(crate) backends: Arc<Backends>,
+    /// The dictionaries configured for this execution.
+    ///
+    /// Populated prior to guest execution, and never modified.
+    pub(crate) dictionaries: Arc<Dictionaries>,
+    /// The dictionaries configured for this execution.
+    ///
+    /// Populated prior to guest execution, and never modified.
+    pub(crate) dictionaries_by_id: Arc<DictionariesById>,
```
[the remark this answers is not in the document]
JakeChampion: This is excellent, I've given this a go in [`b6e5f3f` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/b6e5f3fe3154bfb6f87255f7a0f3dfcd06e72a49) -- all the tests pass 🎉

JakeChampion:
Thanks for the thorough review @cratelyn - I've added the `PrimaryMap` implementation and moved the validation logic out of main, along with using DictionaryName instead of String as the keys in the Dictionaries HashMap

I think this is ready for another round of reviewing now 😄

aturon:
> I think this is ready for another round of reviewing now 😄

Huzzah! I've got the review baton for this next round, and will do that first thing tomorrow morning (Pacific). Thanks again for this work!

lib/src/config/dictionaries.rs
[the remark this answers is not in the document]
JakeChampion: I think I got an implementation working which removes the need for the name field and instead uses `[local_server.dictionaries.name]` 💃 [`76277b3` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/76277b3c46b6f82f2dd3e49a852abbd519868aab)

I had to remove the `impl TryFrom<Table> for Dictionary` as Tables don't have access to their name, what I've done is inlined that code into `impl TryFrom<Table> for DictionariesConfig`

lib/src/wiggle_abi/dictionary_impl.rs
JakeChampion: Memoization was removed in [`5623140` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/5623140308d3e62c9c8b7ae1778b05008efd34a7)

lib/src/config.rs
```diff
@@ -138,15 +159,64 @@ pub struct LocalServerConfig {
...
+    }
+}
+
+/// Configuration settings used for tests.
+///
+/// This represents all of the `fastly.toml` fields whose keys begin with `testing`. Currently this
+/// section of the manifest is only used for providing backend definitions, but additional fields
+/// may be added in the future.
+#[derive(Clone, Debug, Default)]
+pub struct DictionaryConfig {
```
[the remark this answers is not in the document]
JakeChampion: Good spot, I've removed this now 👍

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,143 @@
...
+    use {
+        super::{DictionariesConfig, Dictionary, COUNTER},
+        crate::error::{DictionaryConfigError, FastlyConfigError},
+        crate::wiggle_abi::types::DictionaryHandle,
+        std::{
+            convert::TryFrom,
+            sync::{atomic::Ordering, Arc},
+        },
+        toml::value::{Table, Value},
+    };
```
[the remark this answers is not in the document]
JakeChampion: I think I've now fixed all the import formatting for the imports which were added 🤞

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,143 @@
...
+
+/// A map of [`Backend`] definitions, keyed by their name.
+#[derive(Clone, Debug, Default)]
+pub struct DictionariesConfig(pub HashMap<String, Arc<Dictionary>>);
+
+/// This module contains [`TryFrom`] implementations used when deserializing a `fastly.toml`.
+///
+/// These implementations are called indirectly by [`FastlyConfig::from_file`][super::FastlyConfig],
+/// and help validate that we have been given an appropriate TOML schema. If the configuration is
+/// not valid, a [`FastlyConfigError`] will be returned.
```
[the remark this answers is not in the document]
JakeChampion: The `TryFrom<Table>` implementation for `DictionariesConfig` returns a `Result<Self, FastlyConfigError>` - This is a pattern I came across in `lib/src/config/backends.rs`

If parsing into a `DictionariesConfig` fails then a `DictionaryConfigError` error is returned and eventually mapped into a `FastlyConfigError::InvalidDictionaryDefinition { name: String, err: DictionaryConfigError }` error

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,175 @@
...
+///
+/// A Dictionary consists of a name and an id, but more fields may be added in the future.
+#[derive(Clone, Debug)]
+pub struct Dictionary {
+    // Name must start with alphabetical and contain only alphanumeric, underscore, and whitespace
+    pub name: DictionaryName,
+    pub file: PathBuf,
+}
+
+/// A map of [`Backend`] definitions, keyed by their name.
```
aturon, line 31: Copy-pasted doc comment needs updating.

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -6,23 +6,62 @@ use {
...
+    let obj = json.as_object().expect("Expected the JSON to be an Object");
+    obj.clone()
+}
+
 impl FastlyDictionary for Session {
-    #[allow(unused_variables)] // FIXME: Remove this directive once implemented.
     fn open(&mut self, name: &GuestPtr<str>) -> Result<DictionaryHandle, Error> {
-        Err(Error::NotAvailable("Dictionary lookup"))
+        let name = name.as_str()?.to_owned();
+        self.dictionary_handle(&name)
```
aturon, line 30: FWIW, should be able to avoid copying the name like so:

```suggestion
        self.dictionary_handle(&name.as_str()?)
```

Putting the `&` in front allows Rust to trigger "deref coercion" and turn a [`GuestStr`](https://docs.rs/wiggle/0.29.0/wiggle/struct.GuestStr.html) into a `&str`.

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -6,23 +6,62 @@ use {
...
-    #[allow(unused_variables)] // FIXME: Remove this directive once implemented.
     fn get(
         &mut self,
         dictionary: DictionaryHandle,
         key: &GuestPtr<str>,
         buf: &GuestPtr<u8>,
         buf_len: u32,
     ) -> Result<u32, Error> {
-        Err(Error::NotAvailable("Dictionary lookup"))
+        let key = key.as_str()?.to_owned();
```
aturon, line 40: FWIW, I think you can skip the `to_owned()` here (to avoid a copy), similar to the comment above.

lib/src/error.rs
```diff
@@ -256,6 +279,46 @@ pub enum BackendConfigError {
...
+    #[error("Item key named '{key}' in dictionary named '{name}' is too long, max size is {size}")]
+    DictionaryItemKeyTooLong{ key: String, name: String, size: i32 },
+
+    #[error("The dictionary named '{name}' has too many items, max amount is {size}")]
+    DictionaryCountTooLong{ name: String, size: i32 },
+
+    #[error("Item value named '{key}' in dictionary named '{name}' is too long, max size is {size}")]
+    DictionaryItemValueTooLong{ key: String, name: String, size: i32 },
+}
+
```
aturon, line 321: 😍

lib/src/session.rs
```diff
@@ -67,6 +67,14 @@ pub struct Session {
...
     /// Populated prior to guest execution, and never modified.
     pub(crate) backends: Arc<Backends>,
+    /// The dictionaries configured for this execution.
+    ///
+    /// Populated prior to guest execution, and never modified.
+    pub(crate) dictionaries: Arc<Dictionaries>,
+    /// The dictionaries configured for this execution.
+    ///
+    /// Populated prior to guest execution, and never modified.
+    dictionaries_by_name: PrimaryMap<DictionaryHandle, DictionaryName>,
```
aturon, line 77: Suggestion: I think we can just turn this into:

```rust
dictionaries: PrimaryMap<DictionaryHandle, Dictionary>
```

and load it up as we construct the session below. That would eliminate the need for nested lookup in `dictionary`, and since `Dictionary` contains the name we should still be able to locate it just fine. It would also help clarify the naming here, since `dictionaries_by_name` sounds like a map that keys by name.

wdyt?

aturon (changes requested):
Thanks again for taking this on! I made a review pass and left a few suggestions.

FWIW, I do think your impulse toward memoization was a good one, though I also agree that hot-reloading is nice. I think loading the file on every `get` is OK for now, but we should file a follow-up issue to preload dictionary contents and use file-watching to asynchronously reload on changes. That'll be helpful for people using _really big_ dictionary files.

I didn't have a chance to carefully review the updated TOML loading and schema here. I know @cratelyn has thought more about that so I'll pass the baton back over on that one.

Cheers!

lib/src/session.rs
cratelyn: https://github.com/fastly/Viceroy/pull/61#discussion_r686123479

(_whispering_) this was me!

aturon: Ahha, apologies for the inconsistent direction here!

I'll note that right now the code is preloading the `dictionaries_by_name` map when creating the session, rather than lazily allocating as you open them, which is why it seemed like we could just nix the other map. Either way is fine by me!

cratelyn: Ah! I would tilt towards keeping the map, because it would lead to inconsistent behavior between the production platform and Viceroy. Today, if you call `Dictionary::open("some_name")` twice, you'll be given back two handles with distinct inner `0`, `1` u32 values.

That relates to some platform limits concerning how many dictionaries can be "opened" in a session, so it feels worth keeping the lazy handle-to-name mapping.

aturon: Excellent point @cratelyn. So for clarity, @JakeChampion that means we should move away from the preloading currently being done during `Session` construction, and instead blindly allocate a new handle every time `open` is called (no need for searching by name).

(I could see this behavior changing in C@E eventually, but best to match for now.)

lib/src/wiggle_abi/dictionary_impl.rs
JakeChampion: I couldn't fully avoid the copy in this function for key but, I could limit the copy to only happen when the key does not exist within the dictionary -- [`b6c2e4e` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/b6c2e4ea4b5ff9cc0354095e0544b80a5f1d1bfd)

lib/src/config/dictionaries.rs
JakeChampion: Updated in [`ff17124` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/ff17124d03d8244ad1dfe77bfbaa79e44b55252b)

lib/src/wiggle_abi/dictionary_impl.rs
JakeChampion: That's really nice ☺️ This change was done in [`84661a9` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/84661a9df860000c08fc1e73035c2999edc8b5d3)

lib/src/config.rs
```diff
@@ -55,6 +64,11 @@ impl FastlyConfig {
         &self.local_server.backends.0
     }
 
+    /// Get the backend configuration.
```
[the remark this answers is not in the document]
JakeChampion: Thanks for spotting the copy-paste issues hanging around - fixed in [`5b00fd9` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/5b00fd9c08714300397a23304b8b3d4a13ce4a4f)

lib/src/session.rs
JakeChampion: I've changed this to now allocate handles on demand -- [`edfe5a8` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/edfe5a810dbb1c042fc12b297f6ebb93409105cf)

JakeChampion:
Thanks for the review @aturon - I have addressed all the issues now and I learnt about "deref coercion" ☺️ 

The Dictionary Handles are now created on demand and not reused, to match C@E behaviour 👍 

I also spotted some `.expect("..")` which I left by accident and have converted those to errors under  the `DictionaryConfigError` enum.

I believe this is ready for another round of reviews 🙂

lib/src/config/unit_tests.rs
```diff
@@ -98,17 +149,28 @@ mod testing_config_tests {
             .try_into()
     }
 
-    /// Check that the `testing` section can be deserialized.
+    /// Check that the `local_server` section can be deserialized.
     // This case is technically redundant, but it is nice to have a unit test that demonstrates the
     // happy path for this group of unit tests.
     #[test]
-    fn backend_configs_can_be_deserialized() {
-        static BACKENDS: &str = r#"
```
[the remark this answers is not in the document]
aturon: Ahha, right you are! Good as-is 🙏

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,189 @@
...
+                                    } else {
+                                        Ok(file.into())
+                                    }
+                                }
+                                _ => Err(DictionaryConfigError::InvalidFileEntry),
+                            })?;
+                        check_for_unrecognized_keys(&toml)?;
+                        let dictionary_max_len: usize = 1000;
+                        let dictionary_item_key_max_len: usize = 256;
+                        let dictionary_item_value_max_len: usize = 8000;
```
aturon, line 100: Could we pull these out into a dedicated `limits.rs` file? We're going to want to build more awareness of C@E limits in Viceroy like this, and having them in one place will make life easier for keeping them up to date.

aturon (changes requested):
Thank you, the changes look perfect!

I looked over the TOML deserialization now, and just left one tiny request there. Otherwise, I'm very happy with this PR!

That said, I _believe_ that @cratelyn may have one further though re: the TOML schema (future-facing, thinking about "inline" dictionary definitions), but I'm not sure. I'll pass the baton back over to her to finish this out with you!

lib/src/config/dictionaries.rs
JakeChampion: Done in [`b75566d` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/b75566d5e4e78b7ea8b1e91a409d0aca15f2295a) 👍 

Should these also go in the fastly-shared crate?

lib/src/config/dictionaries.rs
```diff
@@ -0,0 +1,193 @@
...
+                        Ok(Dictionary {
+                            name: name.clone().try_into()?,
+                            file,
+                        })
+                    })
+                    .map_err(|err| FastlyConfigError::InvalidDictionaryDefinition {
+                        name: name.clone(),
+                        err,
+                    })
+                    .map(|def| (def.name.clone(), def))
```
cratelyn, line 168:
```suggestion
                        let name = name.parse()?;
                        Ok((name, Dictionary { file }))
                    })
                    .map_err(|err| FastlyConfigError::InvalidDictionaryDefinition {
                        name: name.clone(),
                        err,
                    })
```

cratelyn: (_see the comments attached to `struct Dictionary`, as well as the `DictionaryName: TryFrom<String>` implementation_)

lib/src/config/unit_tests.rs
```diff
@@ -76,18 +85,60 @@ fn fastly_toml_files_with_simple_backend_configurations_can_be_read() {
...
+            authors = [
+                "Amelia Watson <awatson@fastly.com>",
+                "Inugami Korone <kinugami@fastly.com>",
+            ]
+            language = "rust"
+
+            [local_server]
+                [local_server.dictionaries]
+                    [local_server.dictionaries.a]
+                    file="{}"
```
cratelyn, line 112: oh! i forgot one final detail in my last round of comments.

it would be nice if we were able to future-proof this section, leaving open the possibility of other dictionary formats e.g., defining the dictionary inline inside of the `fastly.toml` itself. (_note: that is absolutely something that we can address another day, since this is already a hefty pr._)

would you mind adding a `format = "json"` field here, and to the `Dictionary` struct? if this is anything _but_ `"json"` we can exit with an error until we introduce support for other formats later.

JakeChampion: That's a great idea, I had a go at this in [`72f860e` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/72f860eb751bd70658e479e3e4d205eeb6020c1f)

lib/src/config.rs
```diff
@@ -138,15 +156,35 @@ pub struct LocalServerConfig {
...
+            BackendsConfig(Default::default())
+        };
+
+        let dictionaries = if let Some(dictionaries) = dictionaries {
+            let dictionaries: Result<DictionariesConfig, FastlyConfigError> =
+                dictionaries.try_into();
+            dictionaries?
+        } else {
+            DictionariesConfig(Default::default())
+        };
```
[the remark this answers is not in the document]
JakeChampion: This is lovely ☺️ 
Added in [`f2687e1` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/f2687e1695d530c02d19251319c666788069e0f7)

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -6,23 +6,58 @@ use {
         session::Session,
         wiggle_abi::{fastly_dictionary::FastlyDictionary, types::DictionaryHandle},
     },
+    std::{
+        convert::{
+            TryFrom,
+            TryInto,
```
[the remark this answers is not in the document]
JakeChampion: You are right, it's gone now 👍

lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -6,23 +6,58 @@ use {
...
     ) -> Result<u32, Error> {
-        Err(Error::NotAvailable("Dictionary lookup"))
+        let key: &str = &key.as_str()?;
+        let dict = self.dictionary(dictionary)?;
+        let file = dict.file.clone();
+        let obj = read_json_file(file);
+        if !obj.contains_key(key) {
+            return Err(Error::UnknownDictionaryItem(key.to_string()));
+        }
+        let item = obj.get(key).unwrap(); // Safe to do due to `!obj.contains_key(&key)` above
```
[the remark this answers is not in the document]
JakeChampion: Nice, I applied this change in [`d2f469f` (#61)](https://github.com/fastly/Viceroy/pull/61/commits/d2f469fa5edb81b17931616cf5e70ad7d0700f4a)

cratelyn (approved):
thank you so much for implementing this @JakeChampion! this is wonderful, wonderful work :sparkles: :heavy_check_mark:

cratelyn:
ah! it looks like the ci problems are due to formatting. `cargo fmt --all` should do the trick!

JakeChampion:
> ah! it looks like the ci problems are due to formatting. `cargo fmt --all` should do the trick!

The formatting issues should be fixed now 👍

aturon:
Looks like Clippy is unhappy. If you run `make ci` locally, that should simulate the CI checks we do, and once that passes we should be good to merge!

JakeChampion:
> Looks like Clippy is unhappy. If you run `make ci` locally, that should simulate the CI checks we do, and once that passes we should be good to merge!

`make ci` now passes for me 👍

aturon:
Looks like there are some dictionary-related test failures on Windows: https://github.com/fastly/Viceroy/pull/61/checks?check_run_id=3355092000

aturon:
My best guess of what's happening:

* TOML deserialization looks for hex escapes like this: https://github.com/alexcrichton/toml-rs/blob/a6420b64d5525d62d07a3b1b7d8e1f9205b91a0f/src/tokens.rs#L388-L391, they start as `\u` or `\U`
* Windows uses `\` in paths
* The temp dir happens to start with `u` or `U` triggering this escape code, and what follows isn't valid hex.

aturon:
The TOML spec indicates we should be using `'` for paths:

> Literal strings are surrounded by single quotes. No escaping is performed so what you see is what you get:
> path = 'C:\Users\nodejs\templates'

Sidenote, this will be an issue for eventual inline dictionaries.

JakeChampion:
Thanks for looking into the issue @aturon -- I've pushed a commit to fix this and can confirm that the tests pass on windows now 🎉 -- https://github.com/JakeChampion/Viceroy/runs/3369942476?check_suite_focus=true

aturon (approved):

aturon:
Passed CI and merged! 🎉 

Thanks for both doing this work, and following through these rounds of review and CI issues. This is an important improvement that will benefit a lot of people!

PR #65, opened 2021-08-23
Add closing functionality to Viceroy for Handles

### LINKED ISSUE #64: Add close functionality to RequestHandle/ResponseHandle and update BodyHandle/StreamingBodyHandle
In order to release a new version of the Rust SDK where we add or update Drop impls for (Request/Response/Body/StreamingBody)Handle so that resources are automatically dropped when out of scope from the Session, we need to update Viceroy and get out a new release so that users don't try to do local compute and then have all of their code break. Especially given how these close calls are automatic and might cause people to be confused if trying out locally vs on the fleet where these changes have all rolled out. This issue is to track that progress. I've already opened up the the corresponding PR needed to the ABI here: https://github.com/fastly/compute-at-edge-abi/pull/6. I've self assigned given the internal work was also done by me and I just want to track this work in our issues.

mgattozzi:
This adds the ability for Viceroy to handle closing handles for
Response, Request, Body, and StreamingBody Handle types via the close
host calls that were added to the public ABI and internally for C@E
which has rolled out to the fleet. While this does not include any SDK
updates for the tests, I've tested this out with the new version of the
Rust SDK to make sure things work out correctly which helped find a bug
or with how I implemented it with Viceroy. With this we can still work
with old versions of the SDK and handle the new SDK as well.

Closes #64

### DIFF
--- a/lib/src/session.rs
+++ b/lib/src/session.rs
@@ -252,6 +252,17 @@ impl Session {
             .ok_or(HandleError::InvalidBodyHandle(handle))
     }
 
+    /// Drop a [`Body`][crate::body::Body] from the [`Session`], given its [`BodyHandle`][crate::wiggle_abi::types::BodyHandle].
+    ///
+    /// Returns a [`HandleError`][crate::error::HandleError] if the handle is not associated with a body in the session.
+    pub fn drop_body(&mut self, handle: BodyHandle) -> Result<(), HandleError> {
+        self.bodies
+            .get_mut(handle)
+            .and_then(Option::take)
+            .map(drop)
+            .ok_or(HandleError::InvalidBodyHandle(handle))
+    }
+
     /// Transition a normal [`Body`][body] into the write end of a streaming body, returning
     /// the original body with the read end appended.
     ///
--- a/lib/src/wiggle_abi/req_impl.rs
+++ b/lib/src/wiggle_abi/req_impl.rs
@@ -458,4 +458,11 @@ impl FastlyHttpReq for Session {
 
         Ok(outcome)
     }
+
+    fn close(&mut self, req_handle: RequestHandle) -> Result<(), Error> {
+        // We don't do anything with the parts, but we do pass the error up if
+        // the handle given doesn't exist
+        self.take_request_parts(req_handle)?;
+        Ok(())
+    }
 }
--- a/lib/src/wiggle_abi/resp_impl.rs
+++ b/lib/src/wiggle_abi/resp_impl.rs
@@ -165,4 +165,11 @@ impl FastlyHttpResp for Session {
         resp.status = status;
         Ok(())
     }
+
+    fn close(&mut self, resp_handle: ResponseHandle) -> Result<(), Error> {
+        // We don't do anything with the parts, but we do pass the error up if
+        // the handle given doesn't exist
+        self.take_response_parts(resp_handle)?;
+        Ok(())
+    }
 }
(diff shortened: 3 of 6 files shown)


lib/src/wiggle_abi/req_impl.rs
```diff
@@ -458,4 +458,11 @@ impl FastlyHttpReq for Session {
 
         Ok(outcome)
     }
+
+    fn close(&mut self, req_handle: RequestHandle) -> Result<(), Error> {
+        //·We·don't·do·anything·with·the·parts,·but·we·do·pass·the·error·up·if
+        //·the·handle·given·doesn't·exist
```
cratelyn, line 464: i think editor whitespace markers got copied in here?

lib/src/wiggle_abi/resp_impl.rs
```diff
@@ -165,4 +165,11 @@ impl FastlyHttpResp for Session {
         resp.status = status;
         Ok(())
     }
+
+    fn close(&mut self, resp_handle: ResponseHandle) -> Result<(), Error> {
+        //·We·don't·do·anything·with·the·parts,·but·we·do·pass·the·error·up·if
+        //·the·handle·given·doesn't·exist
```
cratelyn, line 171: ditto, re: whitespace

lib/src/session.rs
```diff
@@ -252,6 +252,23 @@ impl Session {
             .ok_or(HandleError::InvalidBodyHandle(handle))
     }
 
+    /// Drop a [`Body`][body] from the [`Session`], given its [`BodyHandle`][handle].
```
cratelyn, line 255: i think we can use intradoc links here :slightly_smiling_face:

cratelyn (changes requested):
very tiny changes requested, the core of this all seems good to me!

lib/src/wiggle_abi/req_impl.rs
mgattozzi: They did!

lib/src/session.rs
mgattozzi: Oh right I had just copied from what we had before let me fix that up

lib/src/session.rs
```diff
@@ -252,6 +252,19 @@ impl Session {
...
 
+    /// Drop a [`Body`][crate::body::Body] from the [`Session`], given its [`BodyHandle`][crate::wiggle_abi::types::BodyHandle].
+    ///
+    /// Returns a [`HandleError`][crate::error::HandleError] if the handle is not associated with a body in the session.
+    pub fn drop_body(&mut self, handle: BodyHandle) -> Result<(), HandleError> {
+        self.bodies
+            .get_mut(handle)
+            .and_then(Option::take)
+            // Unlike take_body we don't want to call into_body
+            // as this will cause an InvalidBodyHandle error
```
cratelyn, line 263: sorry, i have one more nitpick.

could you reword this comment slightly? i was confused for a moment why this would cause an invalid handle error; upon further reading i've realized that you meant _this function_ would return said error because `BodyVariant::into_body` will return `None` in the streaming case

mgattozzi:
No worries @cratelyn should be good now

cratelyn (approved):
thank you!!

mgattozzi:
I did a full squash so I can merge with just the one commit @cratelyn and so will need a ♻️ ✔️

cratelyn (approved):
a pr so nice i approved it twice!

mgattozzi:
❤️ Thanks @cratelyn

PR #67, opened 2021-08-24
Add doc/RELEASING.md to document release process

mgattozzi:
I wrote down the release process so that we can just point others to this in the future and have less siloed knowledge around it!


cratelyn (approved):
thank you!! it is wonderful to have this documented somewhere :book:

PR #69, opened 2021-08-30
📑 dictionary lookups can return none

### LINKED ISSUE #68: Fastly crate panicking when a dictionary item doesnt exist
While doing some testing today **locally**, i came across a panic on the fastly crate. This seems to be happening because the dictionary item lookup fails because the item does not exist in the dictionary:
`Aug 27 16:50:20.450 ERROR request{id=0}: Hostcall yielded an error: Unknown dictionary item: Manifest-2d98131
thread 'main' panicked at 'fastly_dictionary::get returned an unrecognized result', /Users/mmohammed/.cargo/registry/src/github.com-1ecc6299db9ec823/fastly-0.7.3/src/dictionary/handle.rs:81:23`

I do not see this error on C@E.

cratelyn:
Fixes #68.

this branch makes some follow-up changes to #61, modifying the hostcall so that we return a `None` status code when a dictionary entry does not exist.


lib/src/wiggle_abi/dictionary_impl.rs
```diff
@@ -4,12 +4,32 @@ use {
...
     },
     std::{convert::TryFrom, fs, path::Path},
     wiggle::GuestPtr,
 };
 
+#[derive(Debug, thiserror::Error)]
+pub enum DictionaryError {
+    /// A dictionary item with the given key was not found.
+    #[error("Unknown dictionary item")]
+    UnknownDictionaryItem,
```
aturon, line 20: Quick question: what's the rationale for dropping the string paylaod here? Viceroy prints out the errors prior to converting them to ABI codes, and it seems helpful to provide that info.

aturon (commented):
Looks good except for one small question!

aturon (approved):
Fab! ✨

cratelyn:
```
  = note: LINK : fatal error LNK1285: corrupt PDB file 'D:\a\Viceroy\Viceroy\target\debug\deps\http_semantics-39397fc60f086e09.pdb'; delete and rebuild
```

i'm guessing latent CI failure, we shall see :recycle:

PR #70, opened 2021-09-03
Improve error when file cannot be read

fgsch:
Use the filename in the output.

### DIFF
--- a/lib/src/config.rs
+++ b/lib/src/config.rs
@@ -72,7 +72,10 @@ impl FastlyConfig {
     /// Parse a `fastly.toml` file into a `FastlyConfig`.
     pub fn from_file(path: impl AsRef<Path>) -> Result<Self, FastlyConfigError> {
         fs::read_to_string(path.as_ref())
-            .map_err(Into::into)
+            .map_err(|err| FastlyConfigError::IoError {
+                path: format!("{}", path.as_ref().display()),
+                err,
+            })
             .and_then(Self::from_str)
     }
 
--- a/lib/src/config.rs
+++ b/lib/src/config.rs
@@ -73,7 +73,7 @@ impl FastlyConfig {
     pub fn from_file(path: impl AsRef<Path>) -> Result<Self, FastlyConfigError> {
         fs::read_to_string(path.as_ref())
             .map_err(|err| FastlyConfigError::IoError {
-                path: format!("{}", path.as_ref().display()),
+                path: path.as_ref().display().to_string(),
                 err,
             })
             .and_then(Self::from_str)
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -222,12 +222,16 @@ pub(crate) enum ExecutionError {
     Instantiation(anyhow::Error),
 }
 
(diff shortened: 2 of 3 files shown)


lib/src/config.rs
```diff
@@ -72,7 +72,10 @@ impl FastlyConfig {
     /// Parse a `fastly.toml` file into a `FastlyConfig`.
     pub fn from_file(path: impl AsRef<Path>) -> Result<Self, FastlyConfigError> {
         fs::read_to_string(path.as_ref())
-            .map_err(Into::into)
+            .map_err(|err| FastlyConfigError::IoError {
+                path: format!("{}", path.as_ref().display()),
```
cratelyn, line 76:
```suggestion
                path: path.as_ref().to_string(),
```

types that implement `Display` automatically implement [`std::string::ToString`](https://doc.rust-lang.org/std/string/trait.ToString.html), so we can use that to get our owned string rather than a `format("{}", ...)` call :slightly_smiling_face:

lib/src/error.rs
```diff
@@ -222,12 +222,16 @@ pub(crate) enum ExecutionError {
     Instantiation(anyhow::Error),
 }
 
-/// Errors that can occur while parsing a `fastly.coml` file.
+/// Errors that can occur while parsing a `fastly.toml` file.
```
cratelyn, line 225: thank you for catching this!

cratelyn (commented):
thank you @fgsch! this is an excellent improvement to our error message.

i have a single, _teensy_ quibble about how we convert our path into a string. other than that, this looks great! :sparkles:

lib/src/config.rs
fgsch: This doesn't work but `path.as_ref().display().to_string()` does.  Is that what you meant?

fgsch:
Somewhat related, I'd like to extend this for `InvalidFastlyToml` error as well. 
I was thinking about copying the from_str implementation into from_file and so I can access the path. 
Do you have any suggestions?

lib/src/config.rs
cratelyn: ah! yes, thank you. i'm mostly concerned about avoiding `format!("{}"` here :slightly_smiling_face:

cratelyn:
> Somewhat related, I'd like to extend this for `InvalidFastlyToml` error as well.
> I was thinking about copying the from_str implementation into from_file and so I can access the path.
> Do you have any suggestions?

for this, i might suggest providing this context elsewhere. threading a `path` through all of the code that isn't otherwise concerned with the file system doesn't feel quite right, so i'd imagine that we'd have better luck logging a warning near the call-site using something like...

```rust
let config = match FastlyConfig::from_file(config_path) {
    Ok(config) => config,
    Err(err) => {
        tracing::warn!(path = ?path, err = %err, "error reading configuration file");
        return Err(err);
    }
}
```

lib/src/config.rs
fgsch: Updated

cratelyn: woohoo!

cratelyn (approved):
this looks good to me!

PR #71, opened 2021-09-06
Default to `./fastly.toml` when config file not specified

### LINKED ISSUE #42: RFC: Reconsider behavior when `-C` is not passed
Currently if you don't provide `-C`, Viceroy will simply not load any configuration.

Normally `-C` is automatically provided by the Fastly CLI, but when invoking Viceroy directly it probably makes sense to default to `fastly.toml` if no manifest is provided. If `fastly.toml` is not found, _then_ we can proceed without configuration.

Thoughts?

kailan:
This closes #42


kailan:
Sorry, did not mean to open this yet. This needs fleshing out as it will break for users that intentionally do not provide a config

PR #75, opened 2021-09-11
Replace hyper-tls with hyper-rustls

### LINKED ISSUE #74: [macOS] Viceroy doesn't support TLS 1.3
On macOS, Viceroy is unable to connect to upstream servers that only accept TLS 1.3.

Trace:

```text
wiggle abi{module="fastly_http_req" function="send"}: h=RequestHandle(1) b=BodyHandle(2) backend=*guest 0x100f5b/24
result=Err(HyperError(hyper::Error(Connect, Error { code: -9836, message: "bad protocol version" })))
Hostcall yielded an error: error trying to connect: bad protocol version
```

An example of such a server is `odoh-target.alekberg.net`.

jedisct1:
`hyper-tls` uses `native-tls`, which behave differently on different platforms.

In particular, TLS 1.3 is not supported on Windows and macOS, which is an issue since more and more servers do not support older versions any more.

`hyper-rustls` uses `rustls`, which works consistently across platforms, and supports TLS 1.3

Fixes #74

### DIFF
--- a/lib/src/upstream.rs
+++ b/lib/src/upstream.rs
@@ -31,12 +31,10 @@ impl Connector {
     }
 }
 
-type BoxError = Box<dyn std::error::Error + Send + Sync>;
-
 impl hyper::service::Service<Uri> for Connector {
     type Response = MaybeHttpsStream<TcpStream>;
     type Error = Box<dyn std::error::Error + Send + Sync>;
-    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, BoxError>> + Send>>;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
 
     fn poll_ready(&mut self, cx: &mut task::Context<'_>) -> Poll<Result<(), Self::Error>> {
         self.https.poll_ready(cx)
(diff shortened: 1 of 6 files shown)


cratelyn:
thank you for opening this @jedisct1!

```
error: the lock file /Users/runner/work/Viceroy/Viceroy/cli/tests/trap-test/Cargo.lock needs to be updated but --locked was passed to prevent this
```

i believe CI is failing due to a rather minor problem, the lockfile in `cli/tests/trap-test` needs updating :slightly_smiling_face:

lib/src/upstream.rs
```diff
@@ -25,15 +26,17 @@ impl Connector {
...
+            https: HttpsConnector::with_native_roots(),
         }
     }
 }
 
+type BoxError = Box<dyn std::error::Error + Send + Sync>;
+
 impl hyper::service::Service<Uri> for Connector {
     type Response = MaybeHttpsStream<TcpStream>;
     type Error = Box<dyn std::error::Error + Send + Sync>;
```
cratelyn, line 38: tiny nit, could we use the `BoxError` alias here as well, just to be consistent? stop me if there's some `rustc` wrinkle preventing that though :innocent:

jedisct1:
Hidden `Cargo.lock` updated :)

cratelyn (approved):
:tada: thank you!!

PR #81, opened 2021-10-14
Introduce a combined async_items table

aturon:
This change to `Session` combines the tables previously used for bodies and pending requests, with the intent to later support async operations that work generically over those types.

Matches a recent C@E implementation change along similar lines.


pchickey (approved):
Looks great!

PR #82, opened 2021-10-19
Support Windows

Integralist:
**NOTE**: I also refactored the workflow to use a 'matrix' so `build-linux` and `build-mac` becomes `build`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         rust-toolchain: [stable]
         os: [ubuntu-latest, macos-latest, windows-latest]
-        arch: [amd64, arm64]
+        arch: amd64
         include:
           - os: ubuntu-latest
             name: linux


.github/workflows/release.yml
```diff
@@ -6,8 +6,21 @@ on:
...
+      matrix:
+        rust-toolchain: [stable]
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        arch: [amd64, arm64]
+        include:
+          - os: ubuntu-latest
+            name: linux
+          - os: macos-latest
+            name: darwin
+            rust-target: x86_64-apple-darwin
```
Integralist, line 20: I _think_ `${{ matrix.rust-target }}` will be an empty string for the other OS' and so I'm presuming that's not a problem when it comes to the `actions-rs/toolchain@v1` step's `target` property, which was only set for macOS.

.github/workflows/release.yml
```diff
@@ -30,47 +44,12 @@ jobs:
           cargo build --all --release --locked
           cd target/release
           strip viceroy
```
jedisct1, line 46: `strip viceroy` may not work on Windows, since the file will be called `viceroy.exe`.

.github/workflows/release.yml
```diff
@@ -17,60 +31,29 @@ jobs:
...
-
-      - name: Build and package
+      - name: Package
         run: |
-          cargo build --all --release --locked
-          cd target/release
-          strip viceroy
-          tar czf viceroy_${{ steps.tagName.outputs.tag }}_darwin-amd64.tar.gz viceroy
+          strip viceroy${{ matrix.extension }}
+          tar czf viceroy_${{ steps.tagName.outputs.tag }}_${{ matrix.name }}-${{ martrix.arch }}.tar.gz viceroy${{ matrix.extension }}
```
jedisct1, line 51: `martrix` -> `matrix`.

Integralist: doh!! 🙂

Integralist: Fixed on both now. Thanks!

PR #88, opened 2021-11-04
GitHub Actions: add aarch64 tarballs for linux and macOS

### LINKED ISSUE #87: Fastly CLI reports error when using is using ARM macOS
✋🏻 Hello

We've had [a report](https://github.com/fastly/cli/issues/458) from @yusukebe that the Fastly CLI is displaying an error to say that a corresponding Viceroy asset cannot be found:

```
ERROR: error downloading latest Viceroy release: no asset found for your OS (darwin) and architecture (arm64).
```

The error is accurate because there is indeed no ARM asset produced for Viceroy.

I realised there was a prior conversation regarding ARM support, see comments from [the Windows support PR](https://github.com/fastly/Viceroy/pull/82#issuecomment-947779171):

```
There are dependencies on C and assembly code (ex: wasmtime-fiber), 
so cross-compilation is not going to work unless you install an additional toolchain.

Long story short, we have to use qemu in order to build the ARM packages.
```

jedisct1:
Fixes #87 and Fixes https://github.com/fastly/cli/issues/458

Note that `macos-11` is not the same as `macos-latest`, and is required to get a recent Xcode version.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,24 +3,34 @@ name: Main
 on:
   push:
     tags:
-      - 'v*.*.*'
+      - "v*.*.*"
 
 jobs:
   build:
     strategy:
       matrix:
         rust-toolchain: [stable]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        arch: [amd64]
+        os: [ubuntu-latest, macos-11, windows-latest]
+        arch: [amd64, arm64]
+        exclude:
+          - os: windows-latest
+            arch: arm64
         include:
           - os: ubuntu-latest
             name: linux
-          - os: macos-latest
+            rust_abi: unknown-linux-gnu
+          - os: macos-11
             name: darwin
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -3,24 +3,34 @@ name: Main
...
+      - "v*.*.*"
 
 jobs:
   build:
     strategy:
       matrix:
         rust-toolchain: [stable]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        arch: [amd64]
+        os: [ubuntu-latest, macos-11, windows-latest]
```
Integralist, line 13: Just to clarify for anyone unfamiliar (such as myself), where the PR description states:

> Note that macos-11 is not the same as macos-latest, and is required to get a recent Xcode version.

It means `macos-11` is actually more recent than `macos-latest`, as documented officially here:

https://github.com/actions/virtual-environments#available-environments

Integralist (approved):
LGTM.

PR #90, opened 2021-11-10
Make URI and Host header semantics more explicit

### LINKED ISSUE #89: Backend requests to TLS origins don't seem to set Host overrides properly
A customer reported that, starting with version 0.42.0 (and untested in a more recent version), host overrides specified via either the `local_server.backends` block of `fastly.toml`... 

```
[local_server]
    [local_server.backends]
        [local_server.backends.www]
            url = "https://plus.stellar.example.com/"
            override_host = "plus.stellar.example.com"
```

...or via the `.with_header` property of `req`...

```
    req.set_pass(true);
        return Ok(req
        .with_path("/health-check")
        .with_header("host","plus.stellar.example.com")
        .send(BACKEND_NAME)?);
        },
    }
```

...will not be set as the `Host` for the request to the backend. Instead, it seems to be preserving the host from the incoming request (i.e. `127.0.0.1:7676`), including scenarios where the client explicitly overrides the `Host` header (i.e. `-H "Host: www.example.com"` with curl). We do seem to set the SNI host properly though. When the origin is a non-TLS host, everything seems to work fine. Also works fine in the XQD environment. 

The customer's origin is an Amazon ELB and they shared the following logs. These have been heavily redacted so the only fields included are `elb_status_code`, `"request"`, and `"domain_name"`, as defined by [Amazon's access log format](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-log-entry-syntax). Note that `"domain_name"` refers to SNI. I've also anonymized their domain: 

1) Failed call to local with no host header but with `override_host` set in local server backend config:
`curl -sv -H 'Authorization: 1234567890' localhost:7676/foo.`
================== ELB log entry ===========================
`404 "GET https://localhost:443/health-check HTTP/2.0" "plus.stellar.example.com" `

2) Successful call on same local using host header:
`curl -sv -H 'Authorization: 1234567890' -H 'host: plus.stellar.example.com' localhost:7676/foo`
================== ELB log entry ===========================
`200 "GET https://plus.stellar.example.com:443/health-check HTTP/2.0" "plus.stellar.example.com"`

The `local_server` section of `fastly.toml` is configured as in the above example. For the first request, the host should have been overridden with `plus.stellar.example.com` but it appears to have been set as `localhost`, as with the original incoming request, but with the port changed out. For the second request, the host header on the original incoming client request appears to have been copied to the backend request. In both cases, the SNI domain was set as expected. 

The first request received a `404`, presumably because the ELB uses Host to match the request to a bucket. 

I've been able to repro this using a Node project on Glitch.com as an origin. The [Node script](https://glitch.com/edit/#!/quixotic-clammy-jeep?path=server.js%3A70%3A5) just echoes your `Host` back to you (if the method is `GET` and the path is `/health-check`) and logs the request on its side. When I used this origin without TLS, everything worked as expected; I saw the expected output on the client side and the request was in my logs on Glitch. When I switched to https for the backend, I didn't see the request logged on the Glitch side, and on the client side, the output was the same as if you connected to Glitch with a Host header that didn't match any service on their side, such as `localhost:7676`. 

If needed, I can get together with someone on Zoom and walk through some repros with the Glitch origin.

aturon:
Previously, we were relying on subtleties of Hyper's behavior to determine the final URI and Host header for outbound requests. Unfortunately, that behavior is sensitive to a number of factors, e.g. whether we happen to be sending a request via an HTTP2 connection.

This commit changes Viceroy to canonicalize the request URI and Host header before it ever reaches Hyper. In most cases this behavior will match what Hyper would do anyway.

Fixes #89

### DIFF
--- a/lib/src/upstream.rs
+++ b/lib/src/upstream.rs
@@ -3,10 +3,11 @@ use crate::{
     wiggle_abi::types::PendingRequestHandle,
 };
 use futures::Future;
-use http::uri;
-use hyper::{client::HttpConnector, Client, Request, Response, Uri};
+use http::{uri, HeaderValue};
+use hyper::{client::HttpConnector, Client, HeaderMap, Request, Response, Uri};
 use hyper_rustls::{HttpsConnector, MaybeHttpsStream};
 use std::pin::Pin;
+use std::str::FromStr;
 use std::task::{self, Poll};
 use tokio::{net::TcpStream, sync::oneshot};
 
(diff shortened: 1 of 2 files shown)


cratelyn (approved):
this was a remarkably tricky bug, and i want to commend you again for tracking down the cause of this @aturon. wonderful work :heavy_check_mark: :sparkles:

PR #94, opened 2021-11-17
#93 improve docs build time

### LINKED ISSUE #93: Immense build time
Hi, first of all, thanks for the great tooling. We use it as a library in our tests. In our POC, we have observed an enormous build time in the Github Actions. Not sure, if it is related to Rust in general or if we can optimize it further.

StarpTech:
This will fix #93

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,10 @@ To install Viceroy as a standalone tool, you'll need to first
 Then run `cargo install viceroy`, which will download and build the latest
 Viceroy release.
 
+## Usage as a library
+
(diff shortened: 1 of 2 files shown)


StarpTech:
Should we ignore markdown files in the action trigger?

cratelyn (approved):
this looks good to me. thank you very much for taking the time to add this documentation @StarpTech! i'm happy we found an answer to fix your build times :running_woman: :zap:

cratelyn:
> Should we ignore markdown files in the action trigger?

i filed #95 to track this; it shouldn't be too difficult to add (_your comment in #93 includes an example after all!_) but we can land your additions here first :slightly_smiling_face:

PR #100, opened 2021-12-01
🎚 Use rustls for https directly, for more control

### LINKED ISSUE #91: Ensure that we don't offer http2 in ALPN
Unfortunately, hyper-rustls's ALPN behavior is based on which Cargo features are enabled. To better match C@E behavior, we should avoid offering ALPN at all. We could do this either by landing a configuration option in hyper-rustls, or by just pulling in the bits of hyper-rustls we need (which are fairly minimal) and customizing the behavior directly.

### LINKED ISSUE #96: Viceroy sending `Transfer-Encoding: chunked` request header with no request body causing 400 errors
With Viceroy, we see it’s sending either `Content-Length: 0` (for non-TLS backend) or `Transfer-Encoding: chunked` (for TLS backend) request header even if there’s no request body set and method is GET, for example.

This seems to be an inconsistent behavior between our real edge vs local instance, and it makes us hard to test TLS backends in some cases. Someone noticed that Google Cloud Run instance refuses such connections (TLS, GET with `Transfer-Encoding: chunked`) and local requests result in 400 errors when testing with Viceroy.

```toml
[local_server]
  [local_server.backends]
    [local_server.backends.backend_1]
      override_host = "httpbin.org"
      url = "https://httpbin.org/"
```

```rs
    let bereq = Request::get("https://httpbin.org/anything");
    let beresp = bereq.send("backend_1").unwrap();

    println!("RESP: {:?} ", beresp);
    Ok(beresp)
```

in the above simple example, I’m using `https://httpbin.org/anything` endpoint so that we can see what headers are sent by Viceroy. And you should see `Transfer-Encoding: chunked` in your curl response when sending a request to your local instance.

https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.3

```
  3.  If a Transfer-Encoding header field is present and the chunked
       transfer coding (Section 4.1) is the final encoding, the message
       body length is determined by reading and decoding the chunked
       data until the transfer coding indicates the data is complete.

       If a Transfer-Encoding header field is present in a response and
       the chunked transfer coding is not the final encoding, the
       message body length is determined by reading the connection until
       it is closed by the server.  If a Transfer-Encoding header field
       is present in a request and the chunked transfer coding is not
       the final encoding, the message body length cannot be determined
       reliably; the server MUST respond with the 400 (Bad Request)
       status code and then close the connection.
```

What’s the best way to mitigate this? Are there any workarounds? It seems Viceroy doesn't work well with Cloud Run backends at least, for this reason.

Thank you

aturon:
Previous work (#75) moved Viceroy to using rustls for TLS, with some good benefits. However, the hyper_rustls library bakes in a fair amount of TLS configuration that we need to control more directly.

Since hyper_rustls is a very thin library, and since we already need our own Hyper connector for other reasons, I opted to just use rustls directly within Viceroy, for maximum clarity/control. The main downside is that there's a bit of boilerplate around TCP/TLS streams.

Along the way, this PR also refactors the CLI backend health checks to use the Viceroy library, rather than connecting in its own way, which will help ensure we keep those in sync.

I've confirmed that this PR:

Fixes #91 
Fixes #96

### DIFF
--- a/lib/src/session.rs
+++ b/lib/src/session.rs
@@ -64,15 +64,15 @@ pub struct Session {
     /// The backends configured for this execution.
     ///
     /// Populated prior to guest execution, and never modified.
-    pub(crate) backends: Arc<Backends>,
+    backends: Arc<Backends>,
     /// The TLS configuration for this execution.
     ///
     /// Populated prior to guest execution, and never modified.
-    pub(crate) tls_config: Arc<rustls::ClientConfig>,
+    tls_config: Arc<rustls::ClientConfig>,
     /// The dictionaries configured for this execution.
     ///
     /// Populated prior to guest execution, and never modified.
-    pub(crate) dictionaries: Arc<Dictionaries>,
+    dictionaries: Arc<Dictionaries>,
     /// The dictionaries configured for this execution.
     ///
     /// Populated prior to guest execution, and never modified.
@@ -80,7 +80,7 @@ pub struct Session {
     /// The path to the configuration file used for this invocation of Viceroy.
     ///
     /// Created prior to guest execution, and never modified.
-    pub(crate) config_path: Arc<Option<PathBuf>>,
+    config_path: Arc<Option<PathBuf>>,
     /// The ID for the client request being processed.
     req_id: u64,
 }
@@ -520,9 +520,14 @@ impl Session {
         self.backends.get(name).map(std::ops::Deref::deref)
     }
 
+    /// Access the backend map.
+    pub fn backends(&self) -> &Arc<Backends> {
+        &self.backends
+    }
+
     // ----- TLS config -----
 
-    /// Reference the TLS configuration.
+    /// Access the TLS configuration.
     pub fn tls_config(&self) -> &Arc<rustls::ClientConfig> {
         &self.tls_config
     }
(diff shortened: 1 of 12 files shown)


lib/src/session.rs
```diff
@@ -65,6 +65,10 @@ pub struct Session {
     ///
     /// Populated prior to guest execution, and never modified.
     pub(crate) backends: Arc<Backends>,
+    /// The TLS configuration for this execution.
+    ///
+    /// Populated prior to guest execution, and never modified.
+    pub(crate) tls_config: Arc<rustls::ClientConfig>,
```
cratelyn, line 71:
```suggestion
    tls_config: Arc<rustls::ClientConfig>,
```

because we mention that this is never modified in the comment, perhaps we can remove the `pub(crate)` visibility here, to help enforce that?

that leaves the `tls_config` accessor below as the sole interface through which other modules can acquire a reference to this :slightly_smiling_face:

cratelyn (approved):
this is wonderful, wonderful work @aturon.

great job :sparkles: :heavy_check_mark:

lib/src/session.rs
cratelyn: (this is one of those "_take it or leave it_" comments)

aturon: oh good catch, i was following the pattern of the other read-only items. they should all be using accessors rather than `pub(crate)`, so I'll add a commit to do that :+1:

aturon:
@cratelyn pushed up a commit tidying up the privacy for read-only session fields.

cratelyn (approved):
:tada: :sparkles: :heart_eyes_cat:

cratelyn:
> @cratelyn pushed up a commit tidying up the privacy for read-only session fields.

@aturon thanks for fixing up the other fields while you were doing this! :slightly_smiling_face:

Issue #103, opened 2021-12-06
rusttls error


arpitdesai:
Customer reported issue while running code on local machine
I'm trying to run the code locally via:
 fastly compute serve
and I keep getting this error:
 ERROR: error during execution process:
 Dec 06 14:32:13.867 ERROR Could not load native certificates: BadDER
 Error: BadCerts(Custom { kind: InvalidData, error: Custom { kind: InvalidData, error: BadDER } }).

aturon:
Quick note: https://github.com/fastly/Viceroy/pull/104 is up, which I hope will alleviate this issue.

PR #104, opened 2021-12-13
🌦 Allow partial CA store to be loaded

aturon:
This change more closely mimics the setup we previously had with hyper_rustls, and allows Viceroy to be used even if only some root certificates could be loaded.

I believe this change will address #103, but we will have to test it to be sure.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -19,7 +19,7 @@ use {
         sync::Arc,
     },
     tokio::sync::oneshot::{self, Sender},
-    tracing::{event, info, info_span, Instrument, Level},
+    tracing::{event, info, info_span, warn, Instrument, Level},
     wasmtime::{Engine, InstancePre, Linker, Module},
 };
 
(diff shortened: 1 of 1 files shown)


lib/src/execute.rs
```diff
@@ -323,8 +323,15 @@ fn configure_tls() -> Result<rustls::ClientConfig, Error> {
     let mut config = rustls::ClientConfig::new();
     config.root_store = match rustls_native_certs::load_native_certs() {
         Ok(store) => store,
-        Err((_, err)) => return Err(Error::BadCerts(err)),
+        Err((Some(store), err)) => {
+            warn!("some certificates could not be loaded: {}", err);
```
cratelyn, line 327:
```suggestion
            warn!(%err, "some certificates could not be loaded");
```

TIOLI :innocent:

cratelyn (dismissed):
:heart_eyes_cat: thank you!!

aturon:
Took your suggestions, re-✅?

cratelyn (approved):
:recycle: :heavy_check_mark: a pr so nice, i approved it twice :v:

PR #111, opened 2022-01-02
Fix local dictionary lookups

thiagopnts:
Right now when running locally with `fastly compute serve`, we're doing look ups against the fastly API dictionary schema instead of the edge dict's key, e.g:
```toml
# given this dictionary settings in fastly.toml, a dict with key "foo" and value "bar"
[local_server.dictionaries.edge_dict]
file = "./dict.json"
format = "json"
```
```rust
// panics locally, works on compute@edge
let _ = fastly::Dictionary::open("edge_dict")
    .get("foo")
    .unwrap();
// works locally, panics on compute@edge
let _ = fastly::Dictionary::open("edge_dict")
    .get("item_key")
    .unwrap();
```

sample project for a quick repro: https://github.com/thiagopnts/fastly-dict-lookup-bug

### DIFF
--- a/lib/src/wiggle_abi/dictionary_impl.rs
+++ b/lib/src/wiggle_abi/dictionary_impl.rs
@@ -54,7 +54,13 @@ impl FastlyDictionary for Session {
         let file = dict.file.clone();
         let obj = read_json_file(file);
         let item = obj
-            .get(key)
+            .get("item_key")
+            .and_then(|item_key| {
+                if item_key == key {
+                    return obj.get("item_value");
+                }
+                None
+            })
             .ok_or_else(|| DictionaryError::UnknownDictionaryItem(key.to_owned()))?;
         let item = item.as_str().unwrap();
         let item_bytes = item.as_bytes();


kailan:
Hi @thiagopnts, thanks for opening an issue.

I believe in this case there is a misunderstanding regarding the format of the dictionary JSON file. Viceroy does not expect the dictionary format that the Fastly API returns, just an object with keys and values.

You can find this documented on the Developer Hub: https://developer.fastly.com/reference/compute/fastly-toml/#dictionaries

I'll close the issue but feel free to re-open if this doesn't answer your question.

thiagopnts:
oh I just tried this now and it works indeed, for some reason I was getting errors about "not adhering to Fastly's API" the first time I tried to use this some time ago and only got it to work once I used the same schema, so I assumed it was required(and I missed that it was fixed/updated? 🤔). Anyways, I'll close this as the plain json k/v works and is much easier to setup, thanks!

kailan:
Glad to hear! That sounds like a confusing error message so if you come across it again feel free to open an issue so we can make it a bit clearer 👍

PR #116, opened 2022-01-11
do not panic on calls to auto_decompress_response_set which are not setting the decompression to gzip

### LINKED ISSUE #115: auto_decompress_response_set has not yet been implemented in Viceroy
Hello, we run into

```
 panicked at 'not implemented: auto_decompress_response_set has not yet been implemented in Viceroy', /home/starptech/.cargo/registry/src/github.com-1ecc6299db9ec823/viceroy-lib-0.2.8/src/wiggle_abi/req_impl.rs:476:9
```

```rs
     // Build client request
        let req = hyper::Request::builder()
            .method(Method::POST)
            .uri("http://example.com")
            .header("authorization", "bearer 123")
            .header("content-type", "application/json")
            .body(Body::from(serde_json::to_string(client_json_body).unwrap()))
            .unwrap();

        // Fire request
        let response = viceroy::handle_request(req, url).await.unwrap();
```

when upgrading from "fastly: 0.8.0" to "fastly: 0.8.1"

It will internally update

```
[[package]]
name = "fastly-shared"
version = "0.8.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "a50fae8c678779e8dcd18e36f9982d09c706764e8976c542fd53193a62d83727"
dependencies = [
 "bitflags",
 "http",
 "thiserror",
]

[[package]]
name = "fastly-sys"
version = "0.8.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "afcf0f870a0c9ac8166b0797a9391e5df81b7b5cc13b53b4d305e0b1368801d4"
dependencies = [
 "bitflags",
 "fastly-shared 0.8.1",
]
```

It was introduced in https://github.com/fastly/Viceroy/commit/fee8165aa6396a348b946a31bc658acb291e8594

Related: https://github.com/fastly/cli/issues/496

JakeChampion:
I've added two tests, one which set the decompression to the default value - this should not cause a panic, and another test which sets the decompression to gzip - this should cause a panic due to it being unimplemented.

This aims to resolve #115

### DIFF
--- a/test-fixtures/src/bin/request.rs
+++ b/test-fixtures/src/bin/request.rs
@@ -9,7 +9,7 @@ use {
         header_append, header_insert, header_remove, header_value_get, method_get, method_set, new,
         uri_get, uri_set, version_get, version_set,
     },
-    fastly_sys::RequestHandle,
+    fastly_sys::{ContentEncodings, RequestHandle},
     http::header::{HeaderName, HeaderValue},
     http::{Method, Uri},
 };
@@ -420,11 +420,18 @@ fn test_header_multi_value_set_and_get() {
     assert_eq!(0, no_header_names.len());
 }
 
+fn test_decompress_gzip_response() {
+    let mut request = FastlyRequestHandle::new();
+
+    request.set_auto_decompress_response(ContentEncodings::GZIP);
+}
+
 fn main() {
     test_version_set_and_get();
     test_uri_set_and_get();
     test_method_set_and_get();
     test_header_value_get_and_insert();
     test_header_append_and_remove();
-    test_header_multi_value_set_and_get()
+    test_header_multi_value_set_and_get();
+    test_decompress_gzip_response()
 }
--- a/test-fixtures/src/bin/request.rs
+++ b/test-fixtures/src/bin/request.rs
@@ -420,10 +420,10 @@ fn test_header_multi_value_set_and_get() {
     assert_eq!(0, no_header_names.len());
 }
 
-fn test_decompress_gzip_response() {
+fn test_default_decompress_response() {
     let mut request = FastlyRequestHandle::new();
 
-    request.set_auto_decompress_response(ContentEncodings::GZIP);
+    request.set_auto_decompress_response(ContentEncodings::default());
 }
 
 fn main() {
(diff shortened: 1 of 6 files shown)


test-fixtures/src/bin/request.rs
```diff
@@ -420,11 +421,31 @@ fn test_header_multi_value_set_and_get() {
...
+
+    let result = panic::catch_unwind(|| {
+        let mut request = FastlyRequestHandle::new();
+
+        request.set_auto_decompress_response(ContentEncodings::GZIP);
+    });
+    if result.is_ok() {
+        panic!("Expected `request.set_auto_decompress_response(ContentEncodings::GZIP);` to cause a panic")
+    }
+}
```
JakeChampion, line 440: This won't work because wasm32-unknown-unknown is panic="abort" by default.

@cratelyn - do you know of a way that we can test panic behavior within the wasm binaries?

JakeChampion: If we can't test panic behavior then we can remove this test for the change.
This commit shows that the other changes and tests in the pull-request do pass on CI -- [`3166b54` (#116)](https://github.com/fastly/Viceroy/pull/116/commits/3166b54860433add1ba7e1cd607069f5ee6b35ba)

cratelyn: > @cratelyn - do you know of a way that we can test panic behavior within the wasm binaries?

short answer: not _really_

slightly more involved answer: yes, but i don't believe that it would be worth the effort in this particular case. we have a fixture called `trap-test`, which is used to check that a panic is handled correctly. _that said_, it is an involved process involving conditional compilation in `viceroy-lib` and a test fixture that must be compiled separately from the other test fixtures.

if you're curious, `trap-test` and `test-fatalerror-config` are two phrases that will yield interesting result when grepping around the repository!

anyways, **tldr;,** we can remove this test; hopefully, we'll implement that feature soon.

JakeChampion: Thank you @cratelyn ☺️ - I will have a read through as I am curious.

I will also remove this test so that this pull-request can pass the tests on CI 👍

cratelyn: :cat: wonderful! we should also be sure to remove the `panic = "unwind"` profiles from `test-fixtures/Cargo.toml`, and then this should be in good shape. thank you for putting this together @JakeChampion! :trophy:

cratelyn (approved):
:sparkles: :heavy_check_mark:

Issue #46, opened 2021-07-15
Package `README.md` with `viceroy`


cratelyn:
See also #41, and #45.

We should package this project's `README.md` with the `viceroy` crate, so that the helpful information contained within is visible to people visiting the project on crates.io! :package:

djKooks:
Hello @cratelyn 
Could you check https://github.com/fastly/Viceroy/pull/118 ?

cratelyn:
hello @djKooks! sure thing, and thank you for opening that PR :smile_cat: :trophy:

cratelyn:
this is now fixed via #118. hooray! :sparkler:

PR #118, opened 2022-01-20
Define README path in config toml

djKooks:
Hello,
I've define path of README in toml, to let document displayed in crate.

It aims to solve the following https://github.com/fastly/Viceroy/issues/46


cratelyn (approved):
thank you!

Issue #76, opened 2021-09-24
Instrument and emit stats on program wall clock duration


cratelyn:
Fastly's UI provides stats on the total wall clock of a Compute@Edge program running in production. Viceroy should emit information about the wall clock duration of a single instance execution, so that users can reason about the behavior of their program in production.

* see also #77.
* see also #78.

cratelyn:
cc: @djKooks, i was guessing based on your activity in #118 and #63 that you were looking for good starter issues in this repository. this issue might be a good one to pick up, if you are interested!

you can see where we currently log how much heap space was used by a Wasm program, here: https://github.com/fastly/Viceroy/blob/main/lib/src/execute.rs#L279-L282

we could use [`Instant::now`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.now) and it's `duration_since` method to find the time elapsed before/after instantiating and running the program. then, a call to [`tracing::info!`](https://docs.rs/tracing/latest/tracing/macro.info.html#examples) should let us emit a log message showing the total wall clock duration. :slightly_smiling_face: 

feel free to ping me if you have any further questions :bowing_woman:

djKooks:
@cratelyn great thanks! I'll look on it 🙏

djKooks:
@cratelyn sorry, I'm middle in research, so want to ask some basic question.

It should be the task to emit the CPU/RAM usage. So before going on, could you let me know how can I visualize (through cli) the stats which is currently being supported?

djKooks:
@cratelyn for now I've saw
```
Jan 28 21:42:57.618  INFO Listening on http://127.0.0.1:7878
Jan 28 21:43:04.845  INFO request{id=0}: handling request GET http://127.0.0.1:7878/
Jan 28 21:43:04.855  INFO request{id=0}: request completed using 1.4 MB of WebAssembly heap
```
by running `Default starter for Rust` in build.

So the purpose for task is to display RAM usage, CPU time in this log?

cratelyn:
> @cratelyn for now I've saw
> 
> ```
> Jan 28 21:42:57.618  INFO Listening on http://127.0.0.1:7878
> Jan 28 21:43:04.845  INFO request{id=0}: handling request GET http://127.0.0.1:7878/
> Jan 28 21:43:04.855  INFO request{id=0}: request completed using 1.4 MB of WebAssembly heap
> ```
> 
> by running `Default starter for Rust` in build.
> 
[...]

PR #121, opened 2022-01-29
Add stat for clock runtime

djKooks:
Following https://github.com/fastly/Viceroy/issues/76
(please ignore branch/commit name...)

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -17,6 +17,7 @@ use {
         path::{Path, PathBuf},
         sync::atomic::AtomicU64,
         sync::Arc,
+        time::Instant
     },
     tokio::sync::oneshot::{self, Sender},
     tracing::{event, info, info_span, warn, Instrument, Level},
@@ -228,7 +229,7 @@ impl ExecuteCtx {
         remote: IpAddr,
     ) -> Result<(), ExecutionError> {
         info!("handling request {} {}", req.method(), req.uri());
-
+        let now = Instant::now();
         let session = Session::new(
             req_id,
             req,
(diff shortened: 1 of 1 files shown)


lib/src/execute.rs
```diff
@@ -228,7 +229,7 @@ impl ExecuteCtx {
         remote: IpAddr,
     ) -> Result<(), ExecutionError> {
         info!("handling request {} {}", req.method(), req.uri());
-
+        let now = Instant::now();
```
cratelyn, line 232: how about renaming this variable ever so slightly, to something like...

```suggestion
        let start_timestamp = Instant::now();
```

lib/src/execute.rs
```diff
@@ -276,11 +277,15 @@ impl ExecuteCtx {
             .expect("`memory` is exported")
             .size(&store);
 
+        let distant = Instant::now().duration_since(now);
```
cratelyn, line 280: ...and similarly, a slight name tweak

```suggestion
        let request_duration = Instant::now().duration_since(start_timestamp);
```

lib/src/execute.rs
```diff
@@ -276,11 +277,15 @@ impl ExecuteCtx {
...
+
         info!(
             "request completed using {} of WebAssembly heap",
             bytesize::ByteSize::kib(heap_pages as u64 * 64)
         );
 
+        info!("request completed in {:?}", distant);
```
cratelyn, line 287: same rename as above, just applying it here so you can fix this with the click of a mouse :computer_mouse: 

```suggestion
        info!("request completed in {:?}", request_duration);
```

cratelyn (commented):
nice work @djKooks! i have just a _small_ pair of nitpicks about variable names. otherwise, this looks great :slightly_smiling_face:

djKooks:
@cratelyn updated 🙏

cratelyn:
this looks like it's almost ready to go @djKooks! there's currently a hiccup in CI here that was addressed in #123. if you rebase this branch on the current `main` we should be able to fix that. (_note: happy to help walk you through that if you run into any questions!_)

djKooks:
@cratelyn got it. Updated~

PR #125, opened 2022-02-08
Implement automatic decompression of responses

### LINKED ISSUE #124: Implement automatic gzip decompression
We recently added a hostcall to C@E for automatically decoding gzipped responses. (See [the SDK here](https://docs.rs/fastly/latest/fastly/http/request/struct.Request.html#method.set_auto_decompress_gzip).)

We need to implement the same functionality in Viceroy so that services using it can be properly exercised in local testing.

(Follow-up to https://github.com/fastly/Viceroy/issues/115.)

acw:
Fixes #124.

The flag marking this use case is held inside the extensions for the request, in the `ViceroyRequestMetadata` structure -- said structure can be extended in the future for other use cases -- and then queried in the final parts of `send_request` in `src/upstream.rs`. The test cases work through an echo server, which just bounces the request right back, which seemed the easiest way to get the various test cases back.

The error test case -- in which the response comes backed marked as gzip-encoded but isn't -- matches the implementation in C@E.


lib/src/session.rs
```diff
@@ -658,6 +658,9 @@ impl Session {
     }
 }
 
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
+pub struct AutoDecompressResponse {}
```
aturon, line 662: This is a cool way to represent the flag, but there is a drawback for the future: we'll need to add additional metadata to requests and responses to keep up with C@E proper. (I'm doing some work that expands those out right now).

I think we should consider more closely following C@E's approach of introducing `FastlyRequestMetadata`/`FastlyResponseMetadata` structs with fields representing the various knobs.

lib/src/body.rs
```diff
@@ -199,6 +206,38 @@ impl HttpBody for Body {
...
+                        }
+                        Poll::Ready(None) => {
+                            match decoder_state.finish() {
+                                Err(e) => return Poll::Ready(Some(Err(e.into()))),
+                                Ok(buffer) => return Poll::Ready(Some(Ok(bytes::Bytes::from(buffer)))),
+                            }
+                        }
+                        Poll::Ready(Some(item)) => {
+                            // put the body back, so we can poll it again next time
+                            self.chunks.push_front(body.into());
```
aturon, line 227: I think we need to push the entire `CompressedHttpBody` chunk to the front of the queue, to continue reading into it.

test-fixtures/src/bin/gzipped-response.rs
```diff
@@ -0,0 +1,68 @@
...
+        .with_header("Content-Encoding", "gzip")
+        .with_body_octet_stream(HELLO_WORLD.as_bytes())
+        .with_auto_decompress_gzip(true)
+        .send(echo_server.clone())?;
+
+    assert_eq!(bad_gzip.get_header_str("Content-Encoding"), Some("gzip"));
+    assert_eq!(bad_gzip.get_content_length(), Some(HELLO_WORLD.len()));
+
+    Ok(())
+}
```
aturon, line 68: This is lovely! As we mentioned yesterday, we'll want to expand out with some uses of `append` operations and streaming as well.

lib/src/body.rs
acw: Ah, yes. Good catch. Should be addressed in 88146fbeba037d351d16729c0b7d03c2ee982a67.

lib/src/session.rs
acw: Done in 7aa1e6d ! Except that I called it `ViceroyRequestMetadata`.

lib/src/upstream.rs
```diff
@@ -164,21 +168,47 @@ pub fn send_request(
...
             .map_err(|e| {
                 eprintln!("Error: {:?}", e);
                 e
-            })?
-            .map(Body::from))
+            })?;
+
+        if try_decompression
+            && basic_response.headers().get(header::CONTENT_ENCODING)
+                == Some(&HeaderValue::from_static("gzip"))
```
aturon, line 197: We should also allow for `x-gzip` per https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#gzip.coding (and the C@E implementation).

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -470,13 +470,25 @@ impl FastlyHttpReq for Session {
...
+        req_handle: RequestHandle,
         encodings: ContentEncodings,
     ) -> Result<(), Error> {
-        if u32::from(encodings) == 1 {
-            unimplemented!("calling auto_decompress_response_set with GZIP has not yet been implemented in Viceroy");
-        } else {
-            Ok(())
+        // NOTE: We're going to hide this flag in the headers of the request in order to decrease
+        // the book-keeping burden inside Session; there already existed a cleaning pass in the
+        // output chain, which we extend for this use case.
```
aturon, line 478: I think this comment is stale.

aturon (changes requested):
Looking really good! Just a couple minor nits!

aturon (approved):
Fantastic work!

Issue #127, opened 2022-02-10
WebAssembly trapped: Exited with i32 exit status 0


AndrewBarba:
When running a basic Swift application Viceroy 0.2.10 produces a misleading trap after a successful request:

<img width="836" alt="Screen Shot 2022-02-10 at 2 49 43 PM" src="https://user-images.githubusercontent.com/1316152/153485748-246f8280-ad75-4983-b848-925f90287241.png">

It says the error code is 0, which to me means a non-error. So I'm not sure if the Swift runtime is doing something incorrectly after it sends the response? Any extra details on the Viceroy side would be helpful and I can handle Swift side.

Application:

```swift
import Compute

@main
struct App {
    static func main() async throws {
        try await onIncomingRequest(handleIncomingRequest)
    }

    static func handleIncomingRequest(req: IncomingRequest, res: OutgoingResponse) async throws {
        try await res.status(200).send("Hello, Swift")
    }
}
```

cratelyn:
thank you for filing this issue @AndrewBarba! you are right that a return code of 0 shouldn't be logged as an error. i've since opened and landed #128, which should address this problem.

the current `main` should no longer emit this warning. i'll wait to close this until a new `viceroy` package has been released :package: :butterfly:

AndrewBarba:
Awesome thanks so much for the quick fix

cratelyn:
this was released in [`0.2.11`](https://crates.io/crates/viceroy/0.2.11), which should be automatically picked up by the CLI :sparkles: i'm going to close this now, thanks again for filing this Andrew!

PR #128, opened 2022-02-10
𝟎 only log non-zero exit status codes

cratelyn:
if a program invokes e.g., `std::process:exit(0)`, this should not cause
confusing messages as reported in #127:

```
  Jan 1 00:00:00.123 ERROR viceroy_lib::execute: WebAssembly huhu: Exited with i32 exit status 0
wasm backtrace:
    0: 0x280ea - <unknown>!__wasi_proc_exit
    1: 0x280f6 - <unknown>!_Exit
```

apologies for the lack of test coverage here.

this is unfortunately difficult to test, because the
`Result<(), ExecutionError>` of `ExecuteCtx::run_guest` is spawned in a
task that we do not join on later. (note: that is intentional)

the alternative of a test that checks for a particular log message felt
brittle, and this is a relatively straightfoward change.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -264,7 +264,10 @@ impl ExecuteCtx {
             .await
             .map(|_| ())
             .map_err(|trap| {
-                event!(Level::ERROR, "WebAssembly trapped: {}", trap);
+                // Be sure that we only log non-zero status codes.
+                if trap.i32_exit_status() != Some(0) {
+                    event!(Level::ERROR, "WebAssembly trapped: {}", trap);
+                }
                 ExecutionError::WasmTrap(trap)
             });
 


lib/src/execute.rs
```diff
@@ -264,7 +264,10 @@ impl ExecuteCtx {
             .await
             .map(|_| ())
             .map_err(|trap| {
-                event!(Level::ERROR, "WebAssembly trapped: {}", trap);
+                // Be sure that we only log non-zero status codes.
+                if trap.i32_exit_status() != Some(0) {
+                    event!(Level::ERROR, "WebAssembly huhu: {}", trap);
```
aturon, line 269: This LGTM except that I don't know what "huhu" is, was it intended?

cratelyn: not intended, thank you for catching this! i've force pushed a fresh commit (_and commit message_) that doesn't include my silly testing string :sweat_smile:

cratelyn:
sigh. i have a sneaking suspicion that [`dac94ad`](https://github.com/fastly/Viceroy/commit/dac94ada133140c476a61469012e6a10a8188a81) is related to the CI failure here.

i don't see this failure when running `cargo clippy` myself, but i'll disclaim that i _am_ running on a linux box. :pensive:

aturon (approved):

PR #132, opened 2022-02-15
🍽 Combine integration test executables for faster builds

acfoltzer:
This moves all of the tests in the top level of `cli/tests` into a single test executable called `integration`. This reduces the number of executables linked when building or rebuilding these tests from 12 to 1. Given the size of `viceroy-lib` and its dependencies, the savings are significant.

Before:

```shell
❯ touch lib/src/lib.rs && time cargo build --tests -j4 -q
cargo build --tests -j4 -q  134.07s user 23.14s system 308% cpu 50.890 total
```

After:

```shell
❯ touch lib/src/lib.rs && time cargo build --tests -j4 -q
cargo build --tests -j4 -q  33.94s user 4.81s system 248% cpu 15.569 total
```

(I ran these with `-j4` to simulate a typical quad core laptop; the difference in wall time is way less dramatic if you have an abundance of cores)

`cargo test` also runs test executables sequentially, so combining the test cases into a single executable exposes additional opportunities for parallelism when running the tests. However the current test harness has a mutex forcing the tests to run sequentially. Lifting this restriction for faster test executions is left for future work 😉


acfoltzer:
This seems to sort of improve CI runtimes, but not as much as I would have expected. The `test` step in Ubuntu went from taking ~26 minutes on typical CI runs to ~18 minutes on this branch, but the Windows and Mac runs seem to have taken the same time. I'm really unsure how to account for this, but since it's not worse, I'm inclined to go ahead with this change and hope to understand those platforms better later.

awortman-fastly (approved):
(some) CI times :chart_with_downwards_trend:

PR #133, opened 2022-02-16
📈 Split `trap-test` into a separate CI job

### LINKED ISSUE #18: Rework CI and trap test for faster check times
Currently, our test for handling traps involves a feature flag, which means CI jobs must compile `viceroy-cli` and `viceroy-lib` twice. We should either parallelize this in a separate job, or consider reworking it to work slightly differently.

acfoltzer:
Fixes #18.

The `trap-test` target requires a different configuration of Cargo features than the normal `test` target. When run in sequence, that means we have to rebuild the entire workspace twice on the same job. By splitting these targets into different CI jobs, they can run in parallel, hopefully reducing the overall wall time of a CI run while keeping the same test coverage.

Along the way, I also made the use of the `cargo --locked` option more consistent in the Makefile, and enforce its presence when using the CI targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
-# Default to using regular `cargo`. The `ci` target will override this.
+# Default to using regular `cargo`. CI targets may override this.
 VICEROY_CARGO=cargo
-export VICEROY_CARGO
 
 .PHONY: format
 format:
--- a/Makefile
+++ b/Makefile
@@ -1,56 +1,67 @@
+# Default to using regular `cargo`. The `ci` target will override this.
+VICEROY_CARGO=cargo
+export VICEROY_CARGO
+
 .PHONY: format
 format:
-	cargo fmt
+	$(VICEROY_CARGO) fmt
 
 .PHONY: format-check
 format-check:
-	cargo fmt -- --check
+	$(VICEROY_CARGO) fmt -- --check
 
 .PHONY: clippy
 clippy:
-	cargo clippy --all -- -D warnings
+	$(VICEROY_CARGO) clippy --all -- -D warnings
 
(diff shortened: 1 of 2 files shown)


acfoltzer:
Before this PR and #132:
image
This PR without #132:
image
This PR rebased to include #132:
image

Makefile
```diff
@@ -1,56 +1,67 @@
+# Default to using regular `cargo`. The `ci` target will override this.
+VICEROY_CARGO=cargo
+export VICEROY_CARGO
```
dcaba, line 3: As I guess we are just replacing this var in the Makefile context (but not being used by child processes that makes the env var a requirement), maybe we don't need to export it?

Makefile
```diff
@@ -1,56 +1,67 @@
+# Default to using regular `cargo`. The `ci` target will override this.
```
dcaba, line 1:
```suggestion
# Default to using regular `cargo`. CI targets may override this.
```
just to be more precise, given now we have two ci entrypoints, and both override it... but maybe this changes in the future (and we may forget to update this)

dcaba (approved):
:clap: Very minor suggestions added, but already LGTM!!

cratelyn (dismissed):
this is stellar work @acfoltzer. thank you for taking the time to address this, it is hugely appreciated.

happy to refresh my approval once @dcaba's comments are addressed. :recycle:

Makefile
acfoltzer: Correct! 87b48ef

cratelyn (approved):
:sparkles:

dcaba (approved):

PR #134, opened 2022-02-16
Add support for writing test cases directly in WAT.

### LINKED ISSUE #130: Add support for running direct WASM / WAT tests directly
Presently, the built and test system assumes that all of the tests are based on Rust code, and live in an appropriate subtree of `test-fixtures`. We should extend this to support hand-written WASM tests, as well, perhaps as a fore-runner to having other languages inside the test suite as well.

acw:
This version leverages `wasmtime`'s ability to compile/run WAT directly to avoid the need to install additional tooling on the build machines, at the cost of requiring some fiddling with the test harness.

Resolves #130.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/cache@v1
       with:
         path: target
-        key: ${{ runner.os }}-cargo-build-target-v1-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ runner.os }}-cargo-build-target-v2-${{ hashFiles('**/Cargo.lock') }}
     - name: Install rustfmt
       run: rustup component add rustfmt
       shell: bash
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,12 +21,12 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/.cargo/registry
-        key: ${{ runner.os }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -29,7 +31,7 @@ jobs:
       uses: actions/cache@v1
       with:
         path: target
-        key: ${{ runner.os }}-cargo-build-target-v1-${{ hashFiles('**/Cargo.lock') }}
+        key: ${{ runner.os }}-cargo-build-target-v${{ env.CACHE_GENERATION }}-${{ hashFiles('**/Cargo.lock') }}
```
fgsch, line 34: Perhaps something to do in a different PR but should this be extended to the other key in that file?

acw: I wasn't sure. It'd be easy enough to do, but I admit I don't have much depth when it comes to these caching mechanisms. So I limited to just the key that seemed to be failing.

Do you understand these things? I'm happy to push a patch if you do and think it'd be a good idea. If not, any idea who does?

fgsch: To some extent :) I think we should cover both. They have the same key (based on the output from the Github Actions) so in theory the jobs can share the files. If they can't (because the keys end up being different) and we need to bump the CACHE_GENERATION value, worse case the first time the job runs it will need to download and build everything. For reference, this will happen naturally if the cache file is not accessed in 7 days.

acw: Done! (In 2b58192)

cratelyn (commented):
this looks great! i love how this new test plumbing looks, and _really_ appreciate a WAT example that we can build other tests from. i caught a tiny comment nitpick related to a previously broken link, but the contents of this PR look good to me. :sparkles:

cratelyn (approved):
who, WAT, when, where, why :sparkles: :heavy_check_mark:

PR #136, opened 2022-02-18
Update cache action and unify steps

fgsch:
While looking at the cache issue from #134, I noticed errors in the
GHA output when creating the cache tarball. Furthermore, a newer
version of the cache action existed with some improvements, for
example the ability to use multiple paths in a single step.

This PR bumps the cache action to v2, merging all the steps together
as with the official example, hopefully fixing the errors and
speeding things up.

It also includes the rustc version hash in the cache key.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,7 @@ jobs:
       with:
         submodules: true
     - name: Install Rust
+      id: toolchain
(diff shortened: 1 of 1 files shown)


cratelyn (approved):
:heart: i am very grateful that you figured this out. thank you @fgsch!

PR #138, opened 2022-02-23
Further improve caching build artifacts

fgsch:
Turns out, the trap-test target creates files under a different
path so use a separate cache for it.


acw (approved):
:tada:

PR #139, opened 2022-03-03
Provide stubs for new framing headers settings

aturon:
This PR adds stub implementations for the new hostcalls providing custom framing header controls. Attempting to use manual controls results in an error for now.


cratelyn (approved):
:framed_picture: :heart_eyes_cat: :heavy_check_mark:

PR #141, opened 2022-03-09
Add `apt-get update` prior to installing Linux cross-compilation

aturon:
This will hopefully address recent failures to publish releases due to aarch64 cross-compilation packages failing to download.


aturon:
cc @jedisct1

pchickey (approved):

Issue #146, opened 2022-04-04
Make environment variables available inside local server


GeeWee:
In our test suite I have a few places I'd like to e.g. toggle some authentication logic based on environment variables.

This currently seems impossible, as Viceroy only exposes the environment variables it has predefined such as `FASTLY_HOSTNAME`, but it is not possible to set your own env variables and have them be accessible inside `fastly compute serve`

GeeWee:
Yeah I absolutely understand why you're a bit hesitant as you wouldn't be able to read the variables inside the production environment.

We actually do use dictionaries reasonably heavily for something similar. Let me perhaps try to outline the situation we're in a little bit.
- We would like to be able to run integration tests using Viceroy, where each test has it's own database key. In a perfect world we'd like these tests to run simultaneously.
- In production scenarios we use an edge dictionary to look up the database key.
- In integration test scenarios we have a local dictionary with a predefined key.

However, I'd like to be able to swap out the contents of my database key in a, preferably easy way, that means I can run parallel tests.
The way I see it there's a few options
- Environment variables. I'd then use those, and if they don't exist, I'd look in the dictionary for the key. That's what I'm proposing here.
- A way to configure or override dictionaries with env variables. This would solve our problem as well, and might be more appetitizing to you as well? As you're not introducing env vars that won't exist in production.
- Using different `fastly.toml` environments such as `fastly.staging.toml`, but if I want to run e.g. 10 or 20 tests simultaneously, that's quite a lot of files I'd have to write in my filesystem

What are your thoughts?

cratelyn:
after a bit more thought, going with environment variables feels like it won't be the workable answer to this. keeping the environment in which a Wasm program runs in parity with production is an important goal for Viceroy, so i'd like for us to look elsewhere.

the `fastly.toml` environments feel like the best fit for this situation, but i hear you that this would involve a lot of filesystem interaction to run tests. using `viceroy_lib` directly in your tests may be a better route that i might propose, if possible.

to that effect, @GeeWee, i opened up #148 to hopefully accommodate this. this should let you pass a string directly into a `FastlyConfig` constructor, without needing to write files to disk. does that possibly help with your problems?

and to help you on your journeys, this should look vaguely like this (_note that today this still involves writing a config to disk, until the branch above lands._)

```rust
use std::collections::HashMap;
use std::path::PathBuf;
use std::sync::Arc;
use viceroy_lib::config::Backend;

fn example_invocation() {
    // Load the wasm module into an execution context
[...]

PR #148, opened 2022-04-11
🍎 config: expose `FastlyConfig::from_str`

cratelyn:
in order to facilitate using `viceroy_lib` in tests, without needing to
write many `fastly.toml` files to the filesystem, we expose the
`from_str` constructor for `FastlyConfig`.

this patch is made with use cases like those described in #146 in
mind. cc: @GeeWee

### DIFF
--- a/lib/src/config.rs
+++ b/lib/src/config.rs
@@ -80,7 +80,7 @@ impl FastlyConfig {
     }
 
     /// Parse a string containing TOML data into a `FastlyConfig`.
-    fn from_str(toml: impl AsRef<str>) -> Result<Self, FastlyConfigError> {
+    pub fn from_str(toml: impl AsRef<str>) -> Result<Self, FastlyConfigError> {
         toml::from_str::<'_, TomlFastlyConfig>(toml.as_ref())
             .map_err(Into::into)
             .and_then(TryInto::try_into)


cratelyn:
```
error: method `from_str` can be confused for the standard trait method `std::str::FromStr::from_str`
  --> lib/src/config.rs:83:5
   |
83 | /     pub fn from_str(toml: impl AsRef<str>) -> Result<Self, FastlyConfigError> {
84 | |         toml::from_str::<'_, TomlFastlyConfig>(toml.as_ref())
85 | |             .map_err(Into::into)
86 | |             .and_then(TryInto::try_into)
87 | |     }
   | |_____^
   |
   = note: `-D clippy::should-implement-trait` implied by `-D warnings`
   = help: consider implementing the trait `std::str::FromStr` or choosing a less ambiguous method name
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait
```

ah! we can already `std::str::FromStr`. :sweat_smile: 

i see two options:
* allow the lint, said trait implementation delegates to this. having an inherent method just makes this a bit more discoverable.
* close this

aturon:
I think we can either just close (the docs.rs page makes `FromStr` _fairly_ prominent), or else add a doc comment on the struct itself mentioning the `from_str` constructor.

cratelyn:
going to close this, 'twas a bit of a misstep on my part. sorry for the noise!

PR #159, opened 2022-05-25
Fix invalid dictionary format error

fgsch:
Update message and correct reported format.


cratelyn (approved):
thank you so much for spotting and fixing this @fgsch! wonderful work. :pray:

PR #161, opened 2022-06-16
Ignore markdown changes in CI

### LINKED ISSUE #95: Ignore markdown changes in CI
it is possible to define our GitHub Actions so that we skip running our CI process when a PR contains only changes to `.md` files.

that looks something like this:

```yaml
name: CI
on:
  push:
    paths-ignore:
      - '**.md'
      - '*.md'
```

this would make documentation updates easier to land in our project. :book: :love_letter:

is2ei:
Fixes #95 

I tested the change in the PR below:
https://github.com/is2ei/Viceroy/pull/3

Ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,7 @@
-on: pull_request
+on:
+  pull_request:
+    paths-ignore:
+      - '**.md'
 name: Test
 env:
   CACHE_GENERATION: 0


cratelyn (approved):
thank you so much for contributing this @is2ei!

PR #164, opened 2022-08-08
Reduce precision of request duration log

acme:
At the moment, a request logs its duration with needlessly high precision:

```
Aug 05 09:44:15.967  INFO request{id=5}: request completed in 363.058968ms
```

This change reduces the precision to a whole number, which is still precise enough to be useful to humans:

```
Aug 05 09:44:15.967  INFO request{id=5}: request completed in 363ms
```


cratelyn (approved):
great idea! thank you :sparkles: :heavy_check_mark:

PR #168, opened 2022-08-16
Extend backend TLS configuration with cert host and SNI

aturon:
To get closer to parity with C@E backend configuration, we add the ability to specify a certificate hostname, distinct from both the host URI or any override host. In addition, backends can now specify whether to use SNI or not; if they do, and the cert hostname is present, it will be used for SNI.

That is slightly less flexible than C@E, which allows distinct SNI and cert hosts. That's not currently possible with the TLS implementation we're using in Viceroy, and the C@E customization is mostly used to _disable_ SNI, which we provide here using a dedicated flag.

The new backend settings are:

* `cert_host` to specify the hostname for certificate checking (and SNI, if enabled)
* `use_sni`, a boolean specifying whether to employ SNI for the backend.

Both settings are optional. If `cert_host` is not specified, the host from the backend URI is used instead. If `use_sni` is not specified, SNI is enabled by default.

Note: this PR does not include tests, which would require quite a bit of scaffolding to do properly. I have, however, tested locally using a real `fastly.toml` provided by a customer, which requires `cert_host` specification to properly reach the backend.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -59,8 +59,10 @@ pub async fn serve(opts: Opts) -> Result<(), Error> {
         }
 
         for (name, backend) in backends.iter() {
-            let client =
-                Client::builder().build(BackendConnector::new(backend, ctx.tls_config().clone()));
+            let client = Client::builder().build(BackendConnector::new(
+                backend.clone(),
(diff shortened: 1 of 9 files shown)


acw (approved):
:sparkle:

PR #170, opened 2022-08-27
keys can contain `.` but can not cannot be named `.` or `..`

JakeChampion:
Keys in the Object Store must follow the following rules:
* Keys can contain any sequence of valid Unicode characters, of length 1-1024 bytes when UTF-8 encoded.
* Keys cannot contain Carriage Return or Line Feed characters.
* Keys cannot start with `.well-known/acme-challenge/`.
* Keys cannot be named `.` or `..`.

The last one was documented correctly in the rustdoc comment but validated incorrectly as we were checking if the key contains a `.` anywhere when we are meant to only mark invalid if the key is exactly `.` or `..`

Checking the object-store codebase we can confirm the validation is meant to be exactly `.` and `..` -- https://github.com/fastly/object-store/blob/d3da1166a69a02df43b769312f95939ba7e566f2/api/src/api/key.rs#L78-L83

We are also missing another validation rule:
>* Keys cannot contain special charaters '[', ']', '*', '?', '#'"

-- https://github.com/fastly/object-store/blob/d3da1166a69a02df43b769312f95939ba7e566f2/api/src/api/key.rs#L92-L100

I can add that into this pull-request or another pull-request if preferred

### DIFF
--- a/cli/tests/integration/object_store.rs
+++ b/cli/tests/integration/object_store.rs
@@ -218,10 +218,10 @@ async fn object_store_bad_key_values() -> TestResult {
         authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
         language = "rust"
         [local_server]
-        object_store.store_one = [{key = "its.me.the.dot.key", data = "This is some data"}]
+        object_store.store_one = [{key = ".", data = "This is some data"}]
     "#;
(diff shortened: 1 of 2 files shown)


mgattozzi (approved):
Thanks for catching my mistake @JakeChampion! Let's do a follow up PR for the other restriction that I missed. You can tag me for review so we can get that in and it might be worth doing a point release.

PR #177, opened 2022-09-14
Bump wasmtime from 0.34.0 to 0.40.1

mgattozzi:
This updates wasmtime to stay up to date with Compute@Edge in terms of
what version of wasmtime is used. The important change here is that
`ModuleLimits` was removed in version 0.35.0. Lots of the limits we
set before now did not matter and so instead we only needed to set a
few of them with instance limits.

PR #183, opened 2022-09-23
Bump wasmtime from 0.34.0 to 0.40.1 to 1.0.0

alexcrichton:
This builds on the work of https://github.com/fastly/Viceroy/pull/177 to do a further bump up to the 1.0.0 release of Wasmtime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,10 +18,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "aho-corasick"
-version = "0.7.18"
+name = "ahash"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


pchickey (changes requested):
I just went chasing this one down too, only to find you already did it!
Here is the set of comments I came up with for the configuration in execute.rs. I double checked these match production. 
```

    const MB: usize = 1024 * 1024;

    let instance_limits = InstanceLimits {
        // This number matches C@E production
        size: 1 * MB,

        // Core wasm programs have 1 memory
        memories: 1,
        // allow for up to 128MiB of linear memory. Wasm pages are 64k
        memory_pages: 128 * (MB as u64) / (64 * 1024),
        // Core wasm programs have 1 table
        tables: 1,
        // Some applications create a large number of functions, in particular
        // when compiled in debug mode or applications written in swift. Every
        // function can end up in the table
        table_elements: 98765,
        // Number of instances: the pool will allocate virtual memory for this many instances,
        // which limits the number of requests which can be handled concurrently.
        count: InstanceLimits::default().count
    };

    config.allocation_strategy(InstanceAllocationStrategy::Pooling {
        strategy: PoolingAllocationStrategy::NextAvailable,
        instance_limits,
    });

```

pchickey (approved):

PR #185, opened 2022-09-30
run integration tests on port 17878 to avoid port conflicts

joeshaw:
If you're already running viceroy on its default 7878 port, several
intergration tests fail because they try to bind to that port.


cratelyn (approved):
thank you! :sparkles:

PR #186, opened 2022-10-03
Add stubs for mTLS info hostcalls

ha0li:
These stubs return errors for now. There is no existing program using these new calls.


cratelyn (approved):
thank you @ha0li! this looks great.

PR #188, opened 2022-10-06
Implement Async IO Select Hostcalls

mgattozzi:
This commit adds the fastly_async_io hostcalls `select` and `is_ready`
to Viceroy. Up to this point we only would do `select` for things like
`PendingRequest`. Now we abstract over an `AsyncItem` which could be a
`PendingRequest`, `StreamingBody`, or `Body` and you can call `select`
over any of these async io operations. Most of this code is to
accomodate these changes while maintaining the behavior of `select` for
`PendingRequest` utilizing these new calls and creating the `AsyncItem`
enum to hold all of these differing types in the same `async_item` map.

We also add an integration test similar to the C@E platform's to make
sure that we implement the behavior correctly. With this we can be
confident that select works the way we intend it too! As a bonus side
effect of these changes we also add async support for hosts in our
testing code in Viceroy which gives us some more flexibility in what we
can accomplish or use!


mgattozzi:
I finally fixed up the PR to work for all items again and the tests should be passing and with it working on the JS SDK side I'm reasonably confident the code works as intended.

lib/src/wiggle_abi/async_io_impl.rs
```diff
@@ -0,0 +1 @@
+
```
aturon, line 1: Looks like an empty file, should be removed.

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -599,26 +604,38 @@ impl FastlyHttpReq for Session {
...
+            .select_impl(
+                pending_req_handles
+                    .as_slice()?
+                    .iter()
+                    .map(|handle| handle.to_owned().into()),
+            )
+            .await? as u32;
+
+        let item =
+            self.take_async_item(pending_req_handles.get(done_index).unwrap().read()?.into())?;
```
aturon, line 619:
```suggestion
            self.take_async_item(pending_req_handles[done_index].read()?.into())?;
```

aturon (changes requested):
Looks great!

However, there aren't tests of the new async_io functionality. We should be able to re-use the C@E tests directly here, so it should be easy to add. While the existing tests give coverage of select over pending requests, I think it's important to have fuller of the new API.

lib/src/wiggle_abi/req_impl.rs
mgattozzi: I thought this would work but turns out `pending_req_handles` is a &GuestPtr<[PendingReqHandle]> and we can't index directly. I tried out some other things but they were either longer or even more complex for little gain and so I think we need to just keep this unfortunately @aturon

mgattozzi:
Update for todays work: I'm still working on it. The C@E tests have some functionality around handling when to accept requests as well as synchronizing that's a bit hard to replicate. I've got something started, but it's slow going for now. Hopefully I'll have something by tomorrow as I don't expect to get it done in the next hour before I clock out today.

lib/src/wiggle_abi/req_impl.rs
aturon: Ah ok, no worries!

mgattozzi:
Figured it all out! Squashed all my WIP commits to make the history cleaner!

mgattozzi:
And updated the code to remove uneeded printlns and make sure we reinsert items using a Drop guard

aturon (approved):
AWESOME WORK! 🎉

Thank you for seeing this through, it'll be great to get it merged.

aturon:
AWESOME WORK! 🎉

Thank you for seeing this through, it'll be great to get it merged.

cli/tests/integration/async_io.rs
```diff
@@ -0,0 +1,228 @@
...
+use hyper::StatusCode;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+use tokio::sync::Barrier;
+
+// For some reason on windows this test fails with the write body portion due to
+// some difference we believe between unix systems and windows when it comes to
+// hyper. We don't believe this means the implementation is wrong. As such we've
+// disabled this test only on windwos.
```
aturon, line 15:
```suggestion
// On Windows, streaming body backpressure doesn't seem to work as expected, either
// due to the Hyper client or server too eagerly clearing the chunk buffer. This issue does
// not appear related to async I/O hostcalls; the behavior is seen within the streaming body
// implementation in general. For the time being, this test is unix-only.
//
// https://github.com/fastly/Viceroy/issues/207 tracks the broader issue.
```

aturon (approved):
🎉 💯

Issue #190, opened 2022-10-13
👏 move from `structopt` to `clap`


cratelyn:
we use `structopt` to handle argument parsing in the CLI facade of Viceroy. today, that functionality exists in `clap`.

see: https://docs.rs/structopt/latest/structopt/#maintenance

> As clap v3 is now out, and the structopt features are integrated into (almost as-is), structopt is now in maintenance mode: no new feature will be added.

let's migrate over when we have time.

PR #195, opened 2022-11-01
👏   Switch from structopt to clap for argument parsing

itsrainy:
Addresses #190. All the tests are passing and I did some additional manual testing around the verbosity flag, since that one changed the most. Everything seems to work the same as before 👯

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,15 +44,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ec8ad6edb4840b78c5c3d88de606b22252d552b55f3a4699fbb10fc070ec3049"
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.65"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


itsrainy:
Note: this has changed the formatting of the help text slightly. I think it is functionally the same, though I do slightly prefer the aesthetics of the structopt help text. If there's anything specific we'd like to keep about it I can poke around a bit and see what options there are for that. The text-wrapping is the main thing that comes to mind for me. I also like that it printed the version.

before
```
viceroy 0.3.2
Fastly
Viceroy is a local testing daemon for Compute@Edge

USAGE:
    viceroy [FLAGS] [OPTIONS] <input>

FLAGS:
    -h, --help          Prints help information
        --log-stderr    Whether to treat stderr as a logging endpoint
        --log-stdout    Whether to treat stdout as a logging endpoint
    -V, --version       Prints version information
    -v                  Verbosity of logs for Viceroy. `-v` sets the log level to DEBUG and `-vv` to TRACE. This option
                        will not take effect if you set RUST_LOG to a value before starting Viceroy

OPTIONS:
    -C, --config <config-path>    The path to a TOML file containing `local_server` configuration
        --profiler <profiler>
        --addr <socket-addr>      The IP address that the service should be bound to

ARGS:
    <input>    The path to the service's Wasm module
```

after
```
Viceroy is a local testing daemon for Compute@Edge

Usage: viceroy [OPTIONS] <INPUT>

Arguments:
  <INPUT>  The path to the service's Wasm module

Options:
      --addr <SOCKET_ADDR>    The IP address that the service should be bound to
  -C, --config <CONFIG_PATH>  The path to a TOML file containing `local_server` configuration
      --log-stdout            Whether to treat stdout as a logging endpoint
      --log-stderr            Whether to treat stderr as a logging endpoint
  -v...                       Verbosity of logs for Viceroy. `-v` sets the log level to DEBUG and `-vv` to TRACE. This option will not take effect if you set RUST_LOG to a value before starting Viceroy
      --profiler <PROFILER>
  -h, --help                  Print help information
```

It also now emits escape codes for boldness and yellowness.

pchickey (dismissed):
:rocket:

itsrainy:
Ah sorry @pchickey I pushed a change to add the version option back in and that dismissed your review 😬. I'm done poking at this :)

pchickey (approved):

PR #199, opened 2022-11-02
Update fastly_status witx definition to include the new limitexceeded variant

JakeChampion:
This variant is returned when an attempt to allocate a resource has exceeded the maximum number of resources permitted. For example, creating too many response handles or makiing too many dictionary lookups.


acfoltzer (approved):
Looks good, although we typically wouldn't merge an ABI update until it is supported across the fleet. In this case we probably can merge sooner as it is backwards compatible with existing SDKs... we should probably have a more explicit set of criteria for handling these types of changes.

PR #203, opened 2022-11-09
🕳 Implement `finish` semantics for streaming bodies

acfoltzer:
Moving forward, streaming bodies must be affirmatively finished when Wasm programs are finished writing to them. This helps prevent unexpected control flows due to error handling causing an incomplete `transfer-encoding: chunked` streaming body from appearing complete to its recipient.

These changes require updates to the integration test fixtures with a prerelease version of the SDK. Once the 0.9.0 Rust SDK is released, we'll follow up with an update to the released version.

### DIFF
--- a/lib/src/body.rs
+++ b/lib/src/body.rs
@@ -1,5 +1,8 @@
 //! Body type, for request and response bodies.
 
+use crate::streaming_body::StreamingBodyItem;
+use crate::Error;
+
 use {
     crate::error,
     bytes::{BufMut, BytesMut},
@@ -36,7 +39,7 @@ pub enum Chunk {
     /// Since the channel yields chunks, this variant represents a *stream* of chunks rather than
     /// one individual chunk. That stream is effectively "flattened" on-demand, as the `Body`
     /// containing it is read.
-    Channel(mpsc::Receiver<Chunk>),
+    Channel(mpsc::Receiver<StreamingBodyItem>),
     /// A version of `HttpBody` that assumes that the interior data is gzip-compressed.
     CompressedHttpBody(DecoderState, hyper::Body),
 }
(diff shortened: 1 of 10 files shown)


acfoltzer:
Note that we are not ready to cut a release including these changes, so we should probably hold off on merging until then. But the changes are ready for review.

cratelyn (approved):
this looks fantastic @acfoltzer, thank you for handling this tricky issue!

acfoltzer:
@cratelyn I prematurely retagged you on this... my plan is to do the 0.9.0 release, update the test fixtures' `Cargo.toml` appropriately, and then merge+release Viceroy.

lib/src/body.rs
```diff
@@ -1,5 +1,8 @@
 //! Body type, for request and response bodies.
 
+use crate::streaming_body::StreamingBodyItem;
+use crate::Error;
+
 use {
     crate::error,
```
cratelyn, line 7: sorry, nit: can we squish this into the `use` below?

cratelyn (approved):
:heavy_check_mark: this looks great! i have just one _tiny_ optional nitpick about imports, take it or leave it. thank you for doing this @acfoltzer!

lib/src/body.rs
acfoltzer: Done in 5f2d111. Do you know if there's a combination of settings that would get `rust-analyzer` to insert imports this way? https://rust-analyzer.github.io/manual.html#auto-import 

I tend to exclusively use the auto-import feature without even looking at what it inserts, which is how I missed this inconsistency.

acfoltzer:
Please hold... I will update the lockfiles once we have 0.9.1 out the door 😞

joeshaw (approved):
👍

PR #206, opened 2022-11-16
object & secret store config naming changes for consistency

joeshaw:
`object_store` becomes `object_stores` for consistency with `backends`
and `dictionaries`.  `path` becomes `file` for consistency with
dictionaries and geolocation config.

An example object store in the fastly.toml file might now look something
like:

```
[local_server]
object_stores.store_one = [{key = "one", file = "/path/to/something"}]
```

or

```
[local_server]
[local_server.object_stores]
store_one = [{key = "one", file = "/path/to/something"}]
```

or

```
[local_server]
[local_server.object_stores]
[[local_server.object_stores.store_one]]
key = "one"
file = "/path/to/something"
```

Backwards compatibility is maintained.

The Secret Store is updated to match the new Object Store conventions, but backwards compatibility is not maintained (because it's brand new and doesn't need to be).

### DIFF
--- a/cli/tests/integration/object_store.rs
+++ b/cli/tests/integration/object_store.rs
@@ -29,6 +29,36 @@ async fn object_store() -> TestResult {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn object_store_backward_compat() -> TestResult {
+    // Previously the "object_stores" key was named "object_store" and
+    // the "file" key was named "path".  This test ensures that both
+    // still work.
+    const FASTLY_TOML: &str = r#"
+        name = "object-store-test"
+        description = "object store test"
+        authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
+        language = "rust"
+        [local_server]
+        object_store.empty_store = []
+        object_store.store_one = [{key = "first", data = "This is some data"},{key = "second", path = "../test-fixtures/data/object-store.txt"}]
+    "#;
+
+    let resp = Test::using_fixture("object_store.wasm")
+        .using_fastly_toml(FASTLY_TOML)?
+        .against_empty()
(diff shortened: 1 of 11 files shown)


cli/tests/integration/object_store.rs
```diff
@@ -9,8 +9,8 @@ async fn object_store() -> TestResult {
         authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
         language = "rust"
         [local_server]
-        object_store.empty_store = []
-        object_store.store_one = [{key = "first", data = "This is some data"},{key = "second", path = "../test-fixtures/data/object-store.txt"}]
+        object_stores.empty_store = []
```
JakeChampion, line 12: should we also allow the previous name?

joeshaw:
Going to work on this again, try to maintain compatibility and move secret store over as well.

cli/tests/integration/object_store.rs
joeshaw: I added backwards compatibility in 1ba490e.

JakeChampion (approved):

Integralist:
👋🏻 @JakeChampion @joeshaw 

I noticed on https://developer.fastly.com/reference/compute/fastly-toml/#local-server it still says `path` and not `file` (as the CLI was updated to use `file` in the toml parsing/lookup). 

Is this just a case of the documentation has been neglected or did something change since this PR (and the CLI PR) were merged?

I just went to the documentation to see how to setup a local object_store and noticed the code in the CLI would fail to parse `path`.

JakeChampion:
> Is this just a case of the documentation has been neglected or did something change since this PR (and the CLI PR) were merged?

We didn't get around to the changes for the developer portal -- `path` is the new name and is staying around 👍

PR #209, opened 2022-11-24
Update to wasmtime 3.0 and enable wasi-nn calls

geekbeast:
The main purpose of this PR was to support wasi-nn calls inside Viceroy. As the necessary changes weren't available until the wasmtime 3.0 release, I also made the updates to to make it compatible with the broader changes in that specific release.

The two big changes that came in were changes to the pattern for PoolingAllocationConfig and how error/traps are handled. There was also one minor change in that it looks like the linker is no longer under the tokio namespace.


pchickey:
I think we'll want to guard the availability of wasi-nn behind a CLI flag, so that (as our testing tool) we don't create the expectation that wasi-nn is available in production

mtr-fastly:
I have implemented enabling experimental modules via flags and refactored the trap test so that it uses the body instead of headers.

lib/src/execute.rs
```diff
@@ -55,18 +56,21 @@ pub struct ExecuteCtx {
     next_req_id: Arc<AtomicU64>,
     /// The ObjectStore associated with this instance of Viceroy
     object_store: Arc<ObjectStore>,
+    // Whether to enable wasi-nn functionality
+    // wasi_modules: HashSet<WasiModule>,
```
mgattozzi, line 60: Is this meant to be commented out? If so we should remove it

lib/src/execute.rs
```diff
@@ -215,37 +221,20 @@ impl ExecuteCtx {
...
-            Ok(resp) => resp,
+            Ok(resp) => (resp, None),
             Err(_) => match guest_handle
                 .await
                 .expect("guest worker finished without panicking")
             {
-                Ok(_) => Response::new(Body::empty()),
+                Ok(_) => (Response::new(Body::empty()), None),
                 Err(ExecutionError::WasmTrap(_e)) => {
+                    eprintln!("There was an error handling the request {}", _e.to_string());
```
mgattozzi, line 231: We should use the following instead of `eprintln!` here
```suggestion
                    event!(Level::ERROR, "There was an error handling the request {}", _e.to_string());
```

mgattozzi (commented):
Hey @geekbeast just one change needed for the error output and then a question on a thing that seemed turned off. This looks great otherwise. If there's a test you can do to show it working that would be great, but given that it's experimental support I'm not going to require it to approve!

lib/src/execute.rs
geekbeast: Yup, it is unnecessary once experimental module host functions are linked.

lib/src/execute.rs
geekbeast: Done.

mgattozzi (approved):
Thanks for taking the time to get all these in!

PR #210, opened 2022-12-01
Secret Store

joeshaw:
This introduces types and hostcalls for the Secret Store, and implements
them in Viceroy, along with configuration to instantiate them.

In Compute@Edge, a Secret Store is an encrypted, read-only key-value
store for sensitive data.  In Viceroy, however, it is a simple
unencrypted in-memory map defined in the `fastly.toml` file in a
manner similar to Object Stores.

At a high level, a Wasm application using the APIs would:
1. Open a secret store by name
2. Get a secret from the store by name
3. Decrypt the secret by calling its `plaintext` method.

In Viceroy, Secret Stores are configured in the same way Object Stores
are:

```toml
[local_server]
  [local_server.secret_store]
    store_one = [{key = "first", data = "This is some secret data"}, {key = "second", path = "/path/to/secret.json"}]

    [[local_server.secret_store.store_two]]
      key = "first"
      data = "This is also some secret data"

    [[local_server.secret_store.store_two]]
      key = "second"
      path = "/path/to/other/secret.json"
```


mgattozzi:
Hey @joeshaw I took a look and this is amazingly well done. I do see it's trying to merge it into the `acf/finish-streaming-bodies` branch right now so we should probably hold off until that's merged and then set this PR to merge into main instead after that's in, otherwise I think these changes have solid comprehensive test coverage and they're passing so I think we should be okay.

joeshaw:
@mgattozzi Sounds good!  Yeah, I wasn't planning this land this until the other branch has landed.  Once that branch is landed, this PR _should_ automatically re-target `main`.

joeshaw:
@mgattozzi With #203 having been merged and the 0.9.1 Rust SDK having been tagged, this should be ready to merge.

joeshaw:
Rebased on `main`, which switched to Wasmtime 4.0.0 and required some tweaks.

PR #217, opened 2023-01-19
Update to Wasmtime 4.0.0

alexcrichton:
This commit upgrades to Wasmtime 4.0.0 which is the latest version of Wasmtime on crates.io. While Wasmtime 5.0.0 is going to be released tomorrow it's probably best to get this in first since there's a number of API changes here.

The main change in Wasmtime 4 is some updates to `wiggle`-based APIs to handle shared memory. This means that previous accessors may sometimes have another failure condition. For viceroy though these failure conditions are always errors, so I've introduced a new `Error::SharedMemory` here to cover this case.


cratelyn (approved):
a pleasure as always, @alexcrichton! thank you :pray:

PR #224, opened 2023-01-31
Return appropriate exit code in run-mode, rather than just 0 or 1

### LINKED ISSUE #223: Plumb WASI exit codes through to the CLI
In run-mode, rather than just exiting with `ExitCode::SUCCESS` or `ExitCode::FAILURE`, capture the return code from WASI and emit that as the exit code. For example, running the following wasm program in run mode should result in viceroy exiting with error code 4:

```wat
(module
  (import "wasi_snapshot_preview1" "proc_exit"
  (func $proc_exit (param i32)))
  (export "_start" (func $_start))
  (memory 10)
  (export "memory" (memory 0))
  
  (func $_start
    (call $proc_exit (i32.const 4))
  )
)
```

 Here is an example of the [wasmtime cli doing a similar thing](https://github.com/bytecodealliance/wasmtime/blob/main/src/commands/run.rs#L214-L246)

itsrainy:
Fixes #223. This adds wasi-common, wasmtime, and libc as explicit dependencies for the cli, but they were already transitive dependencies, so I think that's probably fine?

As a sidenote, I began extending the integration test builder with an `against_run_mode` fn with the goal of testing this, but it started to get a bit involved and I think that might be best to do in a separate PR.


JakeChampion (dismissed):

JakeChampion (dismissed):

joeshaw (approved):
👍

PR #227, opened 2023-02-15
fix: remove dictionary count limit from viceroy

JakeChampion:
this limit can be raised per service on c-at-e. if the limit was raised on a customers service, then the customer could no longer run their application locally with viceroy. by removing he limit in viceroy, customers who have a himit dictionary limit would be able to run their application locally.

I hit this issue in my own project which uses a dictionary containing ~3400 keys


dennismartensson (dismissed):

fgsch (approved):
LGTM

PR #236, opened 2023-03-20
🪀 Add `fastly_backend` interfaces for backend introspection

acw:
This doesn't actually implement any interesting functionality -- it's not clear what pooling would mean within the context of Viceroy -- but it allows programs that use this flag to function properly.

Because this just bit me, I've also removed the bit of code that checks for unknown flags and errors. I think it actually makes more sense for old versions of Viceroy to ignore flags they don't understand. Admittedly, I'm betting that future changes will be like this one (in that they don't change any, or change very little, Viceroy behavior), and not settings which will totally change behavior.

### DIFF
--- a/lib/compute-at-edge-abi/compute-at-edge.witx
+++ b/lib/compute-at-edge-abi/compute-at-edge.witx
@@ -529,6 +529,98 @@
     )
 )
 
+(module $fastly_backend
+    ;; Returns 1 if a backend with this name exists.
+    (@interface func (export "exists")
+        (param $backend string)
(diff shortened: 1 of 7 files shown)


acw:
Well, OK. I was going to argue a bit, but then we discovered that we were a few functions behind in the backend introspection API, and that this was going to cause problems. Which then made me think "well, that means we can't be super tolerant of unknown things in general", so ... OK, agreed.

Also, this PR just got bigger. Sorry for the surprise embiggening.

cratelyn (approved):

PR #245, opened 2023-04-05
feat: Allow local KV Stores to be defined using local_server.kv_stores

JakeChampion:
This updates the fastly.toml parser to make `local.server.kv_stores` be equivalent to `local_server.object_stores`.
This is intending to be a backwards compatible change 👍 

KV Store is the new name for Object Store, which is why this is doing an alias between kv_stores and object_stores.

Future patches will rename the internal implementation and hostcalls from Object Store to KV Store, I've refrained from doing those changes in this pull-request as those would require adding to the witx definitions and I'm not sure what direction we will be taking for that part of the work.


joeshaw (approved):
👍

dennismartensson (approved):
Thank you for doing this work @JakeChampion

PR #250, opened 2023-04-19
First take at implementing PeekableTask

computermouth:
Mostly replaced PendingRequest with PeekableTask, there were a couple spots where it remained, but I think might be ok?

### DIFF
--- a/lib/src/session/async_item.rs
+++ b/lib/src/session/async_item.rs
@@ -1,4 +1,5 @@
 use crate::{body::Body, error::Error, streaming_body::StreamingBody};
+use anyhow::anyhow;
 use futures::Future;
 use futures::FutureExt;
 use http::Response;
@@ -133,8 +134,10 @@ impl<T: Send + 'static> PeekableTask<T> {
                 *self = PeekableTask::Complete(v)
             } else {
                 // todo, not the correct error type
-                *self =
-                    PeekableTask::Complete(Err(Error::GuestError(wiggle::GuestError::PtrOverflow)));
+                *self = PeekableTask::Complete(Err(anyhow!(
+                    "peekable task sender unexpectedly dropped"
+                )
+                .into()));
             }
         }
     }
@@ -143,7 +146,7 @@ impl<T: Send + 'static> PeekableTask<T> {
         match self {
             PeekableTask::Waiting(rx) => rx
                 .await
-                .map_err(|_| Error::GuestError(wiggle::GuestError::PtrOverflow))?,
+                .map_err(|_| anyhow!("peekable task sender unexpectedly dropped"))?,
             PeekableTask::Complete(res) => res,
         }
     }
--- a/lib/src/session/async_item.rs
+++ b/lib/src/session/async_item.rs
@@ -1,4 +1,8 @@
-use crate::{body::Body, streaming_body::StreamingBody, upstream::PendingRequest};
+use crate::{body::Body, error::Error, streaming_body::StreamingBody};
+use futures::Future;
+use futures::FutureExt;
+use http::Response;
+use tokio::sync::oneshot;
 
 /// Represents either a full body, or the write end of a streaming body.
 ///
@@ -8,7 +12,7 @@ use crate::{body::Body, streaming_body::StreamingBody, upstream::PendingRequest}
 pub enum AsyncItem {
     Body(Body),
     StreamingBody(StreamingBody),
-    PendingReq(PendingRequest),
+    PendingReq(PeekableTask<Response<Body>>),
 }
 
 impl AsyncItem {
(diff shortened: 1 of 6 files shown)


lib/src/session/async_item.rs
```diff
@@ -93,4 +97,62 @@ impl AsyncItem {
...
+
+    /// Block until a response is ready.
+    pub async fn await_ready(&mut self) {
+        if let PeekableTask::Waiting(rx) = self {
+            if let Ok(v) = rx.await {
+                *self = PeekableTask::Complete(v)
+            } else {
+                // todo, not the correct error type
+                *self =
+                    PeekableTask::Complete(Err(Error::GuestError(wiggle::GuestError::PtrOverflow)));
```
acfoltzer, line 137: Since this branch represents the other end of the channel closing, that means the task holding the write end has disappeared without writing to it. The ordinary explanations for this are a timeout causing that task to be dropped, or the task erroring out in a way that it fails to write an `Err` to its oneshot sender, which would be a bug.

Viceroy doesn't currently have timeouts on session execution, so I believe the only explanation for ending up in this branch would be a bug. I think it's probably safe to then return an `Error::Other` like this:

```suggestion
                // todo, not the correct error type
                *self =
                    PeekableTask::Complete(Err(anyhow!("peekable task sender unexpectedly dropped").into()));
```

lib/src/session/async_item.rs
```diff
@@ -93,4 +97,62 @@ impl AsyncItem {
...
+            }
+        }
+    }
+
+    pub async fn recv(self) -> Result<T, Error> {
+        match self {
+            PeekableTask::Waiting(rx) => rx
+                .await
+                .map_err(|_| Error::GuestError(wiggle::GuestError::PtrOverflow))?,
```
acfoltzer, line 146: Need to fix up error value here also

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -556,13 +556,13 @@ impl FastlyHttpReq for Session {
 
         // asynchronously send the request
         let pending_req =
-            PendingRequest::spawn(upstream::send_request(req, backend, self.tls_config()));
+            PeekableTask::spawn(upstream::send_request(req, backend, self.tls_config()));
 
         // return a handle to the pending request
-        Ok(self.insert_pending_request(pending_req))
+        Ok(self.insert_pending_request(pending_req.await))
```
acfoltzer, line 562: I think it'd look more natural to put the `.await` directly after the `spawn` on line 559

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -584,43 +584,43 @@ impl FastlyHttpReq for Session {
...
```
acfoltzer, line 590: Another `.await` would look more natural placed directly after the async expression

acfoltzer (changes requested):

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -598,16 +598,19 @@ impl FastlyHttpReq for Session {
...
+                    let resp = self
+                        .take_pending_request(pending_req_handle)?
+                        .recv()
+                        .await?;
+                    let (resp_handle, resp_body_handle) = self.insert_response(resp);
+                    (1, resp_handle, resp_body_handle)
+                }
+                false => (0, INVALID_REQUEST_HANDLE.into(), INVALID_BODY_HANDLE.into()),
+            },
+            None => (0, INVALID_REQUEST_HANDLE.into(), INVALID_BODY_HANDLE.into()),
```
acfoltzer, line 613: Could you also please make `async_item_mut` not return an `Option`, so we don't have to deal with this redundant case?

acfoltzer (changes requested):

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -598,16 +598,19 @@ impl FastlyHttpReq for Session {
...
-                let resp = self
-                    .take_pending_request(pending_req_handle)?
-                    .recv()
-                    .await?;
-                let (resp_handle, resp_body_handle) = self.insert_response(resp);
-                (1, resp_handle, resp_body_handle)
-            }
-            false => (0, INVALID_REQUEST_HANDLE.into(), INVALID_BODY_HANDLE.into()),
+        let outcome = match self.async_item_mut(handle.into())? {
+            Some(item) => match item.is_ready() {
```
acfoltzer, line 602: Thank you for adding the `is_ready()` here. On reflection I'm a bit unsettled that the previous code didn't lead to any test failures. In particular, [`test_poll()`](https://github.com/fastly/Viceroy/blob/main/test-fixtures/src/bin/upstream-async.rs#L72) over here must be getting backend responses fast enough that we aren't able to effectively test that `poll` doesn't block when the response isn't ready. Adding some sleeps to the [test backends](https://github.com/fastly/Viceroy/blob/a38ed29cf69e7559903e0114b19b6ce446cae97c/cli/tests/integration/upstream_async.rs#L12-L25) probably would be enough to expose that behavior; if you're up for it I would feel much better about landing this if we can show that these functions work as expected.

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -598,21 +598,17 @@ impl FastlyHttpReq for Session {
...
-                false => (0, INVALID_REQUEST_HANDLE.into(), INVALID_BODY_HANDLE.into()),
-            },
-            None => (0, INVALID_REQUEST_HANDLE.into(), INVALID_BODY_HANDLE.into()),
-        };
-        Ok(outcome)
+        match self.async_item_mut(handle.into()) {
+            Ok(_) => {
+                let resp = self
+                    .take_pending_request(pending_req_handle)?
+                    .recv()
```
acfoltzer, line 605: Hmm, something is wrong if the tests using `poll` are passing without calling `is_ready()` and instead calling a bare `recv()` here, which would block rather than returning immediately. I'll check this out locally and see if I can find the problem

acfoltzer:
With #252 landed, I've rebased this branch on `main` and fixed the poll implementation on this branch: https://github.com/fastly/Viceroy/compare/acf/peekable-task

@computermouth I'd prefer to land the changes under this PR so we don't lose the history associated with the discovery of the fix, but I also don't want to force-push your branch without asking first. Please let me know soon how you'd like to proceed

acfoltzer (approved):
Alright, we're in good shape now. Thanks for your patience with this!

PR #252, opened 2023-04-25
🗳️ Add synchronization logic for upstream-async tests

acfoltzer:
This allows us to observe that the implementation of `poll()` doesn't block, unlike its counterpart `wait()`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         submodules: true
     - name: Install Rust
-      run: rustup self update && rustup update stable && rustup default stable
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       with:
         submodules: true
     - name: Install Rust
-      run: rustup update stable && rustup default stable
+      run: rustup update --no-self-update stable && rustup default stable
```
acfoltzer, line 20: This is a workaround for a Github Actions windows-2022 image bug; see https://github.com/actions/runner-images/issues/7491

aturon (approved):

PR #253, opened 2023-04-27
Byoung/object store async lookup

computermouth:
Not sure if some of this is overkill or not. Particularly adding `AsyncItem::PendingLookup(PeekableTask<Result<Vec<u8>, ObjectStoreError>>)` rather than converting the object store response to a Response<Body> and cramming it in `AsyncItem::PendingReq(PeekableTask<Response<Body>>)`.

But I thought it'd certainly be possible that some of the attached functionality on the internal side might warrant a different type.

Also @JakeChampion, I made a minor change to the .witx, for `pending_lookup_wait()`


JakeChampion:
I've not reviewed the code, but I thought I leave a comment to let people know that it does in fact work!
I wrote a version of the JS SDK which uses the new hostcalls that this PR has implemented and it is working like a charm.

To run the test JS SDK application, you can use the steps below:
❯ git clone https://github.com/fastly/async-store-lookups-test-app
❯ cd async-store-lookups-test-app/
❯ npm install
❯ VICEROY_PATH="/path/to/my/local/viceroy/which/implements/async/lookup/target/release/viceroy" npm start

The application should now be running on <http://127.0.0.1:7676/>

Visiting <http://127.0.0.1:7676/> should return the word `world` 
Visiting <http://127.0.0.1:7676/readme> should return the contents of the `README.md` file

If the above two routes work - then async lookups are working 🥳

lib/src/error.rs
```diff
@@ -238,6 +238,10 @@ pub enum HandleError {
     #[error("Invalid pending request handle: {0}")]
     InvalidPendingRequestHandle(crate::wiggle_abi::types::PendingRequestHandle),
 
+    /// A lookup handle was not valid.
+    #[error("Invalid pending lookup handle: {0}")]
```
JakeChampion, line 242: Should we add "Object Store" into the error message? It might help people know what the lookup handle is in relation to

lib/src/session/async_item.rs
```diff
@@ -14,6 +15,7 @@ pub enum AsyncItem {
     Body(Body),
     StreamingBody(StreamingBody),
     PendingReq(PeekableTask<Response<Body>>),
+    PendingLookup(PeekableTask<Result<Vec<u8>, ObjectStoreError>>),
```
JakeChampion, line 18: What do you think about aliasing `PeekableTask<Result<Vec<u8>, ObjectStoreError>` as `PendingLookup` and then using that throughout the code?

lib/src/session.rs
```diff
@@ -639,6 +639,52 @@ impl Session {
...
+    }
+
+    /// Take ownership of a [`PendingLookup`], given its [`PendingObjectStoreHandle`].
+    ///
+    /// Returns a [`HandleError`] if the handle is not associated with a pending lookup in the
+    /// session.
+    pub fn take_pending_lookup(
+        &mut self,
+        handle: PendingObjectStoreHandle,
+    ) -> Result<PeekableTask<Result<Vec<u8>, ObjectStoreError>>, HandleError> {
```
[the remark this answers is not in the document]
computermouth: Correct!

lib/src/session/async_item.rs
computermouth: That sounds reasonable to me

lib/src/error.rs
computermouth: That sounds smart, I think I might do a lot of `s/lookup/object/g`, or I suppose now `value` or `kv`. `PendingKVRequest` `PendingKVRequestHandle`???

dennismartensson:
Great to see this making progress! On have not reviews but want to request that we change all references to object store to kv store, since this will ship after the name change has happened.

lib/src/error.rs
JakeChampion: yes, kv for sure 👍

lib/compute-at-edge-abi/typenames.witx
```diff
@@ -94,6 +94,8 @@
 (typename $dictionary_handle (handle))
 ;;; A handle to an Object Store.
 (typename $object_store_handle (handle))
+;;; A handle to a pending Object Store request.
+(typename $pending_object_store_handle (handle))
```
aturon, line 98: We might want to call this `$pending_object_store_lookup_handle`. I realize that's verbose, but it makes room for other async operations on the object story. (And, with the impending rename, it'd be `$pending_kv_lookup_handle` which isn't too bad.)

lib/src/session/async_item.rs
aturon: Can we call this variant `PendingKvLookup` (and rename the session functions accordingly)?

aturon (commented):
Looks good overall, just had a few comments about names to take into account together with the kv renaming.

lib/compute-at-edge-abi/typenames.witx
JakeChampion: good thinking, the store has other methods such as list keys and metadata which would be good to have async when they are implemented 👍

dennismartensson:
Here are the tickets that I think outline all the stuff we should add https://fastly.atlassian.net/browse/EDATA-55

lib/src/session/async_item.rs
computermouth: Done :+1:

lib/src/error.rs
computermouth: Done :+1:

lib/compute-at-edge-abi/typenames.witx
computermouth: Done :+1:

lib/src/session.rs
```diff
@@ -641,48 +641,45 @@ impl Session {
...
     /// Insert a [`PendingLookup`] into the session.
     ///
-    /// This method returns a new [`PendingObjectStoreHandle`], which can then be used to access
+    /// This method returns a new [`PendingKvLookupHandle`], which can then be used to access
     /// and mutate the pending lookup.
-    pub fn insert_pending_lookup(
-        &mut self,
-        pending: PeekableTask<Result<Vec<u8>, ObjectStoreError>>,
-    ) -> PendingObjectStoreHandle {
+    pub fn insert_pending_lookup(&mut self, pending: PendingKvTask) -> PendingKvLookupHandle {
```
aturon, line 646: Can you rename these functions to include `kv` as well?

lib/src/session.rs
```diff
@@ -641,48 +641,45 @@ impl Session {
...
+            .push(Some(AsyncItem::PendingKvLookup(pending)))
             .into()
     }
 
-    /// Take ownership of a [`PendingLookup`], given its [`PendingObjectStoreHandle`].
+    /// Take ownership of a [`PendingLookup`], given its [`PendingKvLookupHandle`].
     ///
     /// Returns a [`HandleError`] if the handle is not associated with a pending lookup in the
     /// session.
     pub fn take_pending_lookup(
```
aturon, line 656: Can you rename these functions to include `kv` as well?

lib/src/session.rs
```diff
@@ -641,48 +641,45 @@ impl Session {
...
-            .ok_or(HandleError::InvalidPendingLookupHandle(handle))
+            .ok_or(HandleError::InvalidPendingKvLookupHandle(handle))
     }
 
-    /// Get a reference to a [`PendingLookup`], given its [`PendingObjectStoreHandle`].
+    /// Get a reference to a [`PendingLookup`], given its [`PendingKvLookupHandle`].
     ///
     /// Returns a [`HandleError`] if the handle is not associated with a lookup in the
     /// session.
     pub fn pending_lookup(
```
aturon, line 674: Can you rename these functions to include `kv` as well?

aturon (approved):
Looks great! ✨

Issue #255, opened 2023-05-03
Use the wasmtime on-demand allocator


acfoltzer:
For `run`: this execution mode is one-and-done and not meant to emulate heap limits and such from real C@E, this should use Wasmtime's default allocator. This will both lower the virtual memory footprint of executions and allow unit tests to use more than the C@E heap limit (currently 128MiB)

For `serve`: [`StoreLimitsBuilder`](https://docs.rs/wasmtime/latest/wasmtime/struct.StoreLimitsBuilder.html) should let us provide fidelity against the real platform limits

PR #257, opened 2023-05-03
🕷️ Rework integration tests to allow parallel test execution

acfoltzer:
No more `TEST_LOCK`! The integration tests now take less than 5 seconds to execute on my workstation, vs 79 seconds previously.

This required two main pieces of work.

## 💽 Run the Viceroy test server on an ephemeral port 

The server spun up to run incoming requests through Wasm now is bound to an ephemeral port rather than being hardcoded to port 7878, and the test requests have their URLs modified to include the ephemeral port assigned to the test.

This change technically means the `Test::against*()` methods can no longer be used to send requests to destinations other than the Viceroy that's been spun up for this test, but that's the only way they're currently used anyway.

The main concrete change here for test authors is that you no longer specify `localhost:7878` in your test requests, you instead just provide a request path.

## 🤵 Redesign test backend servers to also use ephemeral ports 

This is a broader-reaching change in a few ways:

- Backends now _must_ have a service function defined, or test setup will panic. This was already how the integration tests behaved semantically, but this does change some of the syntax around their definitions.

- Test servers for backends are spun up once per `Test`, rather than per invocation of `Test::against*()`. The existing suite of test backends was already mostly stateless, so again not a huge impact on test authors.

- A full URL can no longer be specified for backends. Instead, a path may be provided in order to exercise the backend path prefix behavior of Viceroy.

- Tests that must know dynamically how to reach a backend server now must explicitly start the backend servers before getting a URI suitable for the purpose. This only shows up in dynamic backend tests.

## 🛍️ Grab bag

- The level of parallelism made possible by this change meant that I had to tweak the default number of instances in the Wasmtime pooling allocator down from its default of 1000. I don't expect a limit of 100 to cause anyone problems, but this also raised the possibility of switching away from the pooling allocator for most use cases. A followup issue is available at https://github.com/fastly/Viceroy/issues/255.

- The additional parallelism also caused a massive CPU bottleneck when JIT compiling the test fixture Wasm modules. The main workspace now enables opt-level 1 for debug builds, substantially improving runtime performance of the tests without heavily impacting compile speed.

- The `trap-test` was using a `#[path = "..."]` mechanism to reuse code from the integration test suite. I didn't know this was a mechanism that existed, and it was cool to learn about it. But it also was incompatible with `common` now having a child module, so I inlined the very small handful of definitions that test was reusing.

- I added a couple extra functions to support cases where a test might want to define a backend config separately from its service function, for example if starting a `Test` configuration from a `fastly.toml` file. These currently go unused because none of the existing `fastly.toml`-using tests define any backends.

- The new `TestBackends` interfaces are awfully panicky. I figure this is OK for test infrastructure, but am open to suggestions if folks feel this is unwise. I tried to at least document the various expectations, and make the panic messages relevant beyond just a line number.

- The graceful shutdown logic is no longer strictly necessary unless we write enough tests with enough backends to run out of ports. But because it seems to work fine and is a tidy way to handle things, I left it in.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,8 @@ exclude = [
 # commands like `cargo run` will run the CLI binary by default.
 # See: https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection
 default-members = [ "cli" ]
+
+[profile.dev]
+# Since some of the integration tests involve compiling Wasm, a little optimization goes a long way
+# toward making the test suite not take forever
+opt-level = 1
--- a/cli/tests/integration/async_io.rs
+++ b/cli/tests/integration/async_io.rs
@@ -28,8 +28,7 @@ async fn async_io_methods() -> TestResult {
     let barrier_2 = barrier.clone();
 
     let test = Test::using_fixture("async_io.wasm")
-        .backend("Simple", "http://127.0.0.1:9000/", None)
-        .async_host(9000, move |req: Request<Body>| {
+        .async_backend("Simple", "/", None, move |req: Request<Body>| {
             assert_eq!(req.headers()["Host"], "simple.org");
             let req_count_1 = req_count_1.clone();
             let barrier_1 = barrier_1.clone();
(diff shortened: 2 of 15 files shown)


acfoltzer:
I'm happy to report this also takes our fully-cached CI time from 15m30s to around 5m30s. I suspect the bulk of that is from the opt-level change which took maybe 1% of the overall time I spent on this PR, but still we do love to see it.

cratelyn (approved):
:basketball: :basket: :triumph: slam dunk

this is such a lovely change, i'm delighted.

PR #259, opened 2023-05-10
fix: make is_healthy always return `Unknown` instead of an unsupported error

JakeChampion:
This enables applications to use is_healthy during local development without it erroring, but also without it marking backends and healthy or unhealthy


acfoltzer (approved):
Thank you!

PR #260, opened 2023-05-10
🚧 Add stubs for Cache API primitives

acfoltzer:
This adds stubs to allow programs using the unreleased Cache API to succesfully link in Viceroy. This will be followed later by a mock cache implementation.


JakeChampion (approved):
✨

PR #261, opened 2023-05-11
🎁 prepare release 0.5.0

JakeChampion:
Cutting a new release, this currently contains steps 1 through to 5 of [Releasing Viceroy](https://github.com/fastly/Viceroy/blob/main/doc/RELEASING.md#releasing-viceroy).

> 1. Make sure the Viceroy version has been bumped up to the current release
   version. You might need to bump the minor version (e.g. 0.2.0 to 0.3.0) if
   there are any semver breaking changes. Review the changes since the last
   release just to be sure.
> 1. Update the `Cargo.lock` files by running `make generate-lockfile`.
> 1. Update `CHANGELOG.md` so that it contains all of the updates since the
>    previous version as its own commit.
> 1. Create a local branch in the form `release-x.y.z` where `x`, `y`, and `z` are
>    the major, minor, and patch versions of Viceroy and have the tip of the
>    branch contain the Changelog commit.
> 1. Run `make ci` locally to make sure that everything will pass before pushing
>    the branch and opening up a PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
-## Unreleased
+## 0.5.0 (2023-05-11)
+
+- 🚧 Add stubs for Cache API primitives ([#260](https://github.com/fastly/Viceroy/pull/260))
+- Make is_healthy always return Unknown instead of an unsupporte…
+- 🕷️ Rework integration tests to allow parallel test execution ([#257](https://github.com/fastly/Viceroy/pull/257))
+- Add KVStore async lookup ([#253](https://github.com/fastly/Viceroy/pull/253))
+- Update to Wasmtime 8 ([#251](https://github.com/fastly/Viceroy/pull/251))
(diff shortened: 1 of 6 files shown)


acfoltzer (changes requested):
Gah, I just noticed that there's a semver break with the new `HandleError::InvalidPendingKvLookupHandle` variant. We need to mark that `#[non_exhaustive]` if it's gonna be publicly exposed (this probably applies to some other types as well).

For now, would you mind changing the crate versions to 0.5.0 before we release?

acfoltzer:
Oh, the changes to `insert_pending_request` are also part of the public interface to `viceroy-lib`, so its changes are breaking as well. I think we're probably exposing more than we need to from this crate and laying semver traps for ourselves. I'll make a follow-up issue for us to examine that and hopefully cut a 0.6.0 that makes future changes less disruptive.

acfoltzer (approved):
Thank you! Approved to advance to step 6 🚀

acfoltzer (approved):
Sorry for delay on this, thanks for closing the loop ➰

PR #273, opened 2023-05-31
⛽ -> ⏲️  Switch from fuel to epoch interruptions.

itsrainy:
This PR has two main parts:
- Switching to epoch interruptions (the changes to `lib/src/execute.rs` and `lib/src/linking.rs`)
- Fixing and refactoring the `async_io_methods`, which broke as a result of the switch.

The epoch switch is straightforward: set the config value on the wasmtime Engine and then attach a thread to the ExecuteCtx which increments the epoch every 50 µs. We also added a Drop implementation to `ExecuteCtx` to join that thread.

@jameysharp and I collaborated on chasing down the test issue, namely the `async_io_methods` test had a race condition ([alluded to here](https://github.com/fastly/Viceroy/blob/d3e23d4/test-fixtures/src/bin/async_io.rs#L77-L83)) that required it to sleep for 100ms to give the backends time to work. The switch to epoch interruptions made that test flaky, so I added an additional backend and request to the test fixture as a synchronization mechanism between the guest and the server. Once the server has read enough bytes to relieve the backpressure on the write body, it will complete the sync request, signaling to the guest that the write body should be ready.

### DIFF
--- a/cli/tests/integration/async_io.rs
+++ b/cli/tests/integration/async_io.rs
@@ -84,7 +84,9 @@ async fn async_io_methods() -> TestResult {
                 match req_count_3.load(Ordering::Relaxed) {
                     3 => {
                         // Read at least 4MB and one 8K chunk from the request
-                        // to relieve back-pressure for the guest
+                        // to relieve back-pressure for the guest. These numbers
+                        // come from the amount of data that the guest writes to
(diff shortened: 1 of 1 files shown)


cratelyn (approved):
thank you for doing this :sparkles:

PR #276, opened 2023-06-14
Add a stub for downstream_client_request_id

fgsch:
This PR adds a stub for downstream_client_request_id to avoid crashing Viceroy.

I'm not sure if it'd be better to return something meaningful. Open to suggestions.

### DIFF
--- a/lib/src/wiggle_abi/req_impl.rs
+++ b/lib/src/wiggle_abi/req_impl.rs
@@ -89,6 +89,16 @@ impl FastlyHttpReq for Session {
         }
     }
 
+    #[allow(unused_variables)] // FIXME FGS 2023-06-14: Remove this directive once implemented.
+    fn downstream_client_request_id(
+        &mut self,
+        reqid_out: &GuestPtr<u8>,
+        reqid_max_len: u32,
+        nwritten_out: &GuestPtr<u32>,
+    ) -> Result<(), Error> {
+        Ok(())
+    }
+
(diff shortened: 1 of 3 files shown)


lib/src/wiggle_abi/req_impl.rs
```diff
@@ -89,6 +89,16 @@ impl FastlyHttpReq for Session {
...
     }
 
+    #[allow(unused_variables)] // FIXME FGS 2023-06-14: Remove this directive once implemented.
+    fn downstream_client_request_id(
+        &mut self,
+        reqid_out: &GuestPtr<u8>,
+        reqid_max_len: u32,
+        nwritten_out: &GuestPtr<u32>,
+    ) -> Result<(), Error> {
+        Ok(())
```
aturon, line 99: We should probably at least write 0 to `nwritten_out`.

fgsch: Done.

aturon (approved):
Thanks much for tackling this!

GeeWee:
I realize I'm a little late, but is there no chance we can return a UUID or something that mimicks what you might get in live environments a little closer?

fgsch:
@GeeWee we probably can but do you really need some value or you could get away checking for None?

GeeWee:
I mean I can live with whatever, but currently I generate a UUID for correlation logs, and I use that in e.g. integration test scenarios where I use viceroy to correlate logs from parallel requests. I was hoping to just discard that machinery and switch over to the downstream client ID if possible.

PR #277, opened 2023-06-18
feat: Add a stub for downstream_client_h2_fingerprint

JakeChampion:
This adds a single stub for `downstream_client_h2_fingerprint` which will allow applications using this feature to be able to link in Viceroy

### DIFF
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -152,12 +152,6 @@ fn link_legacy_aliases(linker: &mut Linker<WasmCtx>) -> Result<(), Error> {
     )?;
     linker.alias(
         req,
-        "downstream_client_h2_fingerprint",
-        "env",
-        "xqd_req_downstream_client_h2_fingerprint",
-    )?;
-    linker.alias(
-        req,
         "downstream_client_request_id",
         "env",
         "xqd_req_downstream_client_request_id",
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -152,6 +152,12 @@ fn link_legacy_aliases(linker: &mut Linker<WasmCtx>) -> Result<(), Error> {
     )?;
     linker.alias(
         req,
+        "downstream_client_h2_fingerprint",
+        "env",
+        "xqd_req_downstream_client_h2_fingerprint",
+    )?;
+    linker.alias(
+        req,
         "downstream_client_request_id",
         "env",
         "xqd_req_downstream_client_request_id",
--- a/lib/src/wiggle_abi/req_impl.rs
+++ b/lib/src/wiggle_abi/req_impl.rs
@@ -89,6 +89,16 @@ impl FastlyHttpReq for Session {
         }
     }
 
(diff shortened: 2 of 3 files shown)


lib/src/linking.rs
```diff
@@ -150,6 +150,12 @@ fn link_legacy_aliases(linker: &mut Linker<WasmCtx>) -> Result<(), Error> {
         "env",
         "xqd_req_downstream_client_ip_addr",
     )?;
+    linker.alias(
+        req,
+        "downstream_client_h2_fingerprint",
+        "env",
+        "xqd_req_downstream_client_h2_fingerprint",
+    )?;
```
acfoltzer, line 158: New features should be binding against the official name, we shouldn't be introducing new legacy names
```suggestion
```

acfoltzer (changes requested):

lib/src/linking.rs
JakeChampion: @acfoltzer do you happen to know what the new official name is?

The name in the Rust SDK is `downstream_client_h2_fingerprint` -- https://docs.rs/fastly/latest/src/fastly/http/request/handle.rs.html#890

acfoltzer: The official name is what the witx translates to, so `fastly_http_req::downstream_client_h2_fingerprint` (or however you would spell it in your local environment). The `xqd_*` names are leftover from before we used witx, and are kept around only for compatibility with modules compiled against very old SDKs.

That's all to say that accepting the suggestion to delete this particular change will take care of this entirely 🙂

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -89,6 +89,19 @@ impl FastlyHttpReq for Session {
...
+        &mut self,
+        h2fp_out: &GuestPtr<'a, u8>,
+        h2fp_max_len: u32,
+        nwritten_out: &GuestPtr<u32>,
+    ) -> Result<(), Error> {
+        Err(Error::NotAvailable("Client H2 fingerprint"))
+    }
+
+    #[allow(unused_variables)] // FIXME FGS 2023-06-14: Remove this directive once implemented.
+
```
JakeChampion, line 104:
```suggestion
```

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -89,6 +89,18 @@ impl FastlyHttpReq for Session {
...
+    fn downstream_client_h2_fingerprint<'a>(
+        &mut self,
+        h2fp_out: &GuestPtr<'a, u8>,
+        h2fp_max_len: u32,
+        nwritten_out: &GuestPtr<u32>,
+    ) -> Result<(), Error> {
+        Err(Error::NotAvailable("Client H2 fingerprint"))
+    }
+
+    #[allow(unused_variables)] // FIXME FGS 2023-06-14: Remove this directive once implemented.
```
JakeChampion, line 103:
```suggestion
```

acfoltzer (approved):
lgtm, thank you!

PR #280, opened 2023-06-21
⏱️ Add cross-platform ability to profile guest code in run mode

itsrainy:
This uses the GuestProfiler that is now available in wasmtime 10.0.0 to create a profile in a location specified by the user. Right now we currently just have it working for run-mode, but are planning on adding similar functionality in serve mode, likely creating one profile per-request.

To use this, run the following:

```
viceroy run --profile-guest mymodule.wasm
```

This will result in a profile being saved to the default location (`guest-profile.json`), which you can upload and view on https://profiler.firefox.com.


lib/src/execute.rs
```diff
@@ -403,6 +421,22 @@ impl ExecuteCtx {
...
+        {
+            if let Err(e) = std::fs::File::create(&path)
+                .map_err(anyhow::Error::new)
+                .and_then(|output| profile.finish(std::io::BufWriter::new(output)))
+            {
+                eprintln!("failed writing profile at {}: {e:#}", path.display());
+            } else {
+                eprintln!();
+                eprintln!("Profile written to: {}", path.display());
+                eprintln!("View this profile at https://profiler.firefox.com/.");
```
Integralist, line 436: **Newbie question:** why are we writing what looks to be a success message to stderr?

lib/src/execute.rs
```diff
@@ -361,7 +367,12 @@ impl ExecuteCtx {
         outcome
     }
 
-    pub async fn run_main(self, program_name: &str, args: &[String]) -> Result<(), anyhow::Error> {
+    pub async fn run_main(
+        self,
+        program_name: &str,
+        args: &[String],
+        guest_profile_path: Option<&PathBuf>,
```
JakeChampion, line 374: Question: Does this change, introducing a new parameter to a public function, mean we will need to release this as a new major version of the crate?

Integralist: I'm presuming "no" because it's an additive change to the exposed binary interface (i.e. a new optional flag has been added) but if the user doesn't provide that flag the existing functionality continues to work as expected.

lib/src/execute.rs
jameysharp: I had done it that way in the equivalent code in Wasmtime to reduce the chance that this message might get mixed up with regular output from the wasm guest. I don't feel strongly about whether Viceroy should do the same, where guest output is less likely to be significant.

PR #282, opened 2023-06-28
Fill downstream_client_request_id in

fgsch:
Instead of returning an empty string, use the request id to mimic prod closer. 
Prompted by @GeeWee in https://github.com/fastly/Viceroy/pull/276.


acfoltzer (approved):
looks good, thank you!

PR #283, opened 2023-07-11
😯 Add support for the new secret from_bytes extension.

acw:


test-fixtures/src/bin/secret-store.rs
```diff
@@ -20,4 +21,9 @@ fn main() {
         None => {}
         _ => panic!(),
     }
+
+    // FIXME: uncomment this when we release this API more fully
+    //let hello_bytes = "hello, wasm_world!".as_bytes().to_vec();
+    //let secret = Secret::from_bytes(hello_bytes).unwrap();
+    //assert_eq!("hello, wasm_world!", secret.plaintext());
```
acfoltzer, line 28: Time to FIXME?

acw: As soon as the 0.9.5 release goes out, I'll update it :smile:

acfoltzer: Oh, we have ourselves a circular dependency here, don't we

acfoltzer (approved):

test-fixtures/src/bin/secret-store.rs
acw: :(

acw: Yeah, we can either merge this as is, with a commented-out test case (I can say, I ran it manually and it worked), or we can merge a broken test case and have the SDK release fix it. I think I'd lean towards the former, and then update Viceroy afterwards.

JakeChampion (approved):

Issue #47, opened 2021-07-15
Align port defaults


aturon:
The Fastly CLI defaults to port 7676, while the Viceroy CLI defaults to port 7878.

PR #287, opened 2023-07-12
Change default port from 7878 to 7676, which is what the Fastly CLI defaults to

JakeChampion:
I think this would be viewed as a breaking change as we are changing the default behaviour of the CLI.
If that is the case, then this involves bumping from 0.6.0 to 0.7.0 when it comes to releasing this

Resolves https://github.com/fastly/Viceroy/issues/47

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ built by `fastly compute build`. The Fastly CLI should put the blob at
 viceroy bin/main.wasm
 ```
 
-This will start a local server (by default at: `http://127.0.0.1:7878`), which can
+This will start a local server (by default at: `http://127.0.0.1:7676`), which can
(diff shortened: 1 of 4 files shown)


aturon (approved):
Very much in favor!

README.md
```diff
@@ -49,7 +49,7 @@ built by `fastly compute build`. The Fastly CLI should put the blob at
 viceroy bin/main.wasm
 ```
 
-This will start a local server (by default at: `http://127.0.0.1:7878`), which can
+This will start a local server (by default at: `http://127.0.0.1:7676`), which can
```
Integralist, line 52: Just to confirm that yes the CLI uses port 7676 by default:
https://github.com/fastly/cli/blob/main/pkg/commands/compute/serve.go#L78

PR #288, opened 2023-07-17
switch default log level to "error", add -v to run

joeshaw:
The previous default log level was "off", which meant that errors
encountered during startup with `viceroy run` were not printed and the
process silently failed.  For example, this might happen if you had an
invalid path in your `fastly.toml` file.

In addition, there was no way to turn on verbose log output with
`viceroy run`, only in serve mode.

This commit moves the default log level to "error", and moves the `-v`
flag into the common set.

Making "error" the default meant that processes that called `proc_exit`
would print WASM stacktraces as errors on exit, but this isn't really an
error from the user's perspective, so we check for that (`I32Exit`) and
suppress the error message in that case only.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -54,14 +54,22 @@ pub async fn main() -> ExitCode {
     let cmd = opts.command.unwrap_or(Commands::Serve(opts.serve));
     match cmd {
         Commands::Run(run_args) => {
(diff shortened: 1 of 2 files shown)


joeshaw:
@itsrainy I tagged you in as a potential reviewer since you were the last person to recently touch this.  If you could take a look I'd appreciate it!

PR #296, opened 2023-08-03
Support the new config store hostcalls.

acw:
These work by naively just forwarding the new hostcalls into the underlying dictionary system; these pass the tests, but I'd love someone to confirm (preferably someone named @acfoltzer) that this is intended behavior.


ulyssa (approved):
This looks good to me! And I believe that forwarding is the appropriate behaviour for Viceroy. I believe the hostcall's intended difference in xqd is that it doesn't go through the sidecar, which doesn't matter here.

PR #300, opened 2023-08-04
Use a ResourceLimiter for tracking allocations

elliottt:
Instead of counting the number of pages present in the memory exported by a wasm module, count the memory allocation diffs over time using the `ResourceLimiter` api.

This change helps set us up for component support, as there's no analogue of `get_memory` for components.

### DIFF
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -13,11 +13,42 @@ use {
     wasmtime_wasi_nn::WasiNnCtx,
 };
 
+#[derive(Default)]
+pub struct Limiter {
+    /// Total memory allocated so far.
+    pub memory_allocated: usize,
+}
+
+impl wasmtime::ResourceLimiter for Limiter {
+    fn memory_growing(
+        &mut self,
+        current: usize,
+        desired: usize,
+        _maximum: Option<usize>,
+    ) -> anyhow::Result<bool> {
+        // Track the diff in memory allocated over time. As each instance will start with 0 and
+        // gradually resize, this will track the total allocations throughout the lifetime of the
+        // instance.
+        self.memory_allocated += desired - current;
+        Ok(true)
+    }
+
+    fn table_growing(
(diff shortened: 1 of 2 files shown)


lib/src/linking.rs
```diff
@@ -13,11 +13,42 @@ use {
...
+#[derive(Default)]
+pub struct Limiter {
+    /// Total memory allocated so far.
+    pub memory_allocated: usize,
+}
+
+impl wasmtime::ResourceLimiter for Limiter {
+    fn memory_growing(
+        &mut self,
+        current: usize,
```
acfoltzer, line 25: Can we confirm that this starts at zero? In other words, is `memory_growing` called when the initial heap for an instance is set up?

elliottt: The interface for allocating new `Memory` structures in wasmtime is [`Memory::new_dynamic`](https://github.com/bytecodealliance/wasmtime/blob/4f1b38f0223843bc48a1f6b1a0af0e9f154d00a0/crates/runtime/src/memory.rs#L672) and [`Memory::new_static`](https://github.com/bytecodealliance/wasmtime/blob/4f1b38f0223843bc48a1f6b1a0af0e9f154d00a0/crates/runtime/src/memory.rs#L689). Both dispatch to [`Memory::limit_new`](https://github.com/bytecodealliance/wasmtime/blob/4f1b38f0223843bc48a1f6b1a0af0e9f154d00a0/crates/runtime/src/memory.rs#L723), which will call `memory_growing` with a [`current` value of `0`](https://github.com/bytecodealliance/wasmtime/blob/4f1b38f0223843bc48a1f6b1a0af0e9f154d00a0/crates/runtime/src/memory.rs#L792).

acfoltzer: Awesome, thank you for checking!

acfoltzer (approved):
Looks good, thank you!

PR #301, opened 2023-08-08
Add guest-profiler support to serve mode

jameysharp:


jameysharp:
I don't understand why the `package-check` CI job is failing for me, but since it's failing to pull in `test-fixtures/data/ca.pem` which was just merged a few hours ago, maybe @acw can take a look?

lib/src/execute.rs
```diff
@@ -364,12 +392,7 @@ impl ExecuteCtx {
         outcome
     }
 
-    pub async fn run_main(
-        self,
-        program_name: &str,
-        args: &[String],
-        guest_profile_path: Option<&PathBuf>,
-    ) -> Result<(), anyhow::Error> {
+    pub async fn run_main(self, program_name: &str, args: &[String]) -> Result<(), anyhow::Error> {
```
JakeChampion, line 395: Note: We need to release this as 0.7.x since we are changing the signature of a [public crate function](https://docs.rs/viceroy-lib/0.6.1/viceroy_lib/struct.ExecuteCtx.html#method.run_main)

JakeChampion (approved):

PR #305, opened 2023-08-28
🤝 Honor backend CA certificates

acw:
This should work for both static certificates in the TOML file, presuming you can get them in in a reasonable format, and through dynamic backends.

I've added some test cases, which required adding a `serial_test` dependency so that the existing client cert test doesn't stomp over the execution requirement. I figured it was nice to leave this functionality in, as there may be more complicated situations in which just using a dynamic backends flag was inconvenient.


lib/src/wiggle_abi/req_impl.rs
```diff
@@ -267,6 +267,30 @@ impl FastlyHttpReq for Session {
             "http"
         };
 
+        let ca_certs =
+            if (scheme == "https") && backend_info_mask.contains(BackendConfigOptions::CA_CERT) {
```
fgsch, line 271: As mentioned elsewhere to @acw , none of the other parameters check the scheme.
Do we want to add it to the other places, possibly in a separate PR?

lib/src/upstream.rs
```diff
@@ -115,17 +116,27 @@ impl hyper::service::Service<Uri> for BackendConnector {
         // the future for establishing the TCP connection. we create this outside of the `async`
         // block to avoid capturing `http`
         let connect_fut = self.http.call(backend.uri.clone());
+        let mut custom_roots = self.tls_config.roots.clone();
+        let (added, ignored) = custom_roots.add_parsable_certificates(&self.backend.ca_certs);
```
fgsch, line 120: I don't believe we should be adding this cert to the roots, but rather using it in isolation.

iximeow: what isolation are you looking for? `custom_roots` is a copy of the root cert store for this invocation of the `Connector`, so the additional `self.backend.ca_roots` is only present in the custom root CA set provided for this specific connection attempt

iximeow: (i'd misunderstood the original comment, "isolation" here means "if there's a CA cert, it should be the only cert used in a root store for validation" rather than the behavior in this PR of "use the CA cert, and the system roots, as acceptable roots for the remote certificate")

lib/src/wiggle_abi/req_impl.rs
acw: Since it's a bit out of scope for the PR, we can possibly do this as a cleanup pass later.

fgsch: 👍 wfm.

lib/src/upstream.rs
fgsch: IOW, if we cannot parse the cert we will just proceed with no cert? Is that correct?

acw: That's correct!

fgsch (dismissed):
👍

JakeChampion:
> This should work for both static certificates in the TOML file, presuming you can get them in in a reasonable format,

Is there an example of what this reasonable format should look like? Also, would it be possible to supply a filepath instead of in-lining into the TOML file?

JakeChampion:
@acw would it be possible to have some tests added to `/lib/src/config/unit_tests.rs` which test out the new `fastly.toml` fields? 🤞

acw:
> Is there an example of what this reasonable format should look like?

Only in the test cases. I could put an example config somewhere, but I'm not sure where? Do we have a standard place to put examples?

acw:
@JakeChampion Now with tests, which can be showed as examples, and support for file references. Also did arrays, while I was in there, so you can add multiple certs.

JakeChampion (approved):

PR #307, opened 2023-09-12
🌭 Update ABI definitions and stub out some hostcalls

acfoltzer:
Adds several new hostcalls and type definitions to the ABI. The level of implementation varies:

- `fastly_http_body::abandon` is fully implemented.

- `fastly_http_req::send_v2` and friends will work, but do not yet populate the new `$error_detail` field with information about the failure.

- `fastly_http_req` hostcalls `downstream_client_oh_fingerprint`, `redirect_to_websocket_proxy_v2`, `redirect_to_grip_proxy_v2`, and `fastly_key_is_valid` are stubs that return `Error::NotAvailable`.

Beyond these added functions, there are a number of formatting fixes in the witx to make the Viceroy copy of these files more precisely match what we use internally to define the platform.

### DIFF
--- a/lib/compute-at-edge-abi/compute-at-edge.witx
+++ b/lib/compute-at-edge-abi/compute-at-edge.witx
@@ -58,7 +58,18 @@
         (result $err (expected $num_bytes (error $fastly_status)))
     )
 
+    ;;; Frees the body on the host.
+    ;;;
+    ;;; For streaming bodies, this is a _successful_ stream termination, which will signal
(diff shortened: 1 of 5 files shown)


acw (approved):
Thank you for doing this!

I'm going to approve, but I'm a little sad that we've standardized on

```
    word))
```

instead of

```
    word
  )
)
```

since the latter makes for smaller diffs. but it's not a big deal.

PR #308, opened 2023-09-15
📞 Support the gRPC flag for dynamic backends

acw:
Enabling this flags adds **experimental** support for gRPC origins for compute services.


acfoltzer (approved):
Looks good; let's go ahead and merge with an update to follow for the test case following the SDK release

lib/src/config/backends.rs
```diff
@@ -126,13 +127,26 @@ mod deserialization {
...
+                .remove("grpc")
+                .map(|grpc| {
+                    if let Value::Boolean(grpc) = grpc {
+                        Ok(grpc)
+                    } else {
+                        Err(BackendConfigError::InvalidGrpcEntry)
+                    }
+                })
+                .transpose()?
+                .unwrap_or(true);
```
acfoltzer, line 140: @acw is this turning gRPC on by default for static backends?

acw: crud. yes.

PR #311, opened 2023-09-18
Reset the default value for `grpc` for static backends to false.

acw:


acfoltzer (approved):
In the subsequent PR to land tests for this new flag, can we include a test to avoid a regression?

acfoltzer (approved):
♻️

PR #313, opened 2023-09-19
👻 Add options to customize behavior of unknown Wasm imports

acfoltzer:
Adds a new shared option to the CLI with three behavior modes:

```
      --unknown-import-behavior <UNKNOWN_IMPORT_BEHAVIOR>
          Set the behavior for unknown imports.

          Note that if a program only works with a non-default setting for this flag, it is unlikely to be publishable to Fastly.

          [default: link-error]

          Possible values:
          - link-error:   Unknown imports are rejected at link time (default behavior)
          - trap:         Unknown imports trap when called
          - zero-or-null: Unknown imports return zero or a null pointer, depending on the type
```

This ended up touching more files than I'd expect, because I had to make the integration test harness bubble out link-time errors rather than `.expect()`ing them, so there are quite a few new `?`s across the integration test suite.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2385,6 +2385,7 @@ dependencies = [
  "bytes",
  "bytesize",
  "cfg-if",
+ "clap",
  "cranelift-entity 0.88.2",
  "fastly-shared",
  "flate2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ opt-level = 1
 [workspace.dependencies]
 anyhow = "1.0.31"
 base64 = "0.21.2"
+clap = { version = "^4.0.18", features = ["derive"] }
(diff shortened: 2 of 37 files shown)


tedmielczarek-fastly:
This sounds great! Since we're using the privileged hostcall `bill_for_service` in the challenge service, we currently have a cargo `viceroy` feature to avoid referencing that hostcall for local testing. I think the `zero-or-null` behavior would work perfectly there and make local testing less annoying.

iximeow (approved):
.... it occurs to me that this is a fascinating way to discover at runtime if a program's dependencies really do need browser-shaped wasm glue, or if maybe you'd never actually reach any of that at runtime. hmmmm..

very handy!

awortman-fastly (approved):
i should approve from the right account though

PR #314, opened 2023-09-25
Lower Hostcall error log level to DEBUG

awilliams-fastly:
The Fastly CLI executes Viceroy with a [`-v`](https://github.com/fastly/cli/blob/24761dcccd81d16e3873c1fa65e03fd83486442f/pkg/commands/compute/serve.go#L534) argument, which corresponds to `INFO` log level. This change drops the log level of hostcall errors to `DEBUG`, so that they're not displayed to users running apps via `fastly compute serve`.

As an example, `BufferLengthError` is a sentinel error used to indicate to the caller that the passed in buffer was [too small](https://github.com/fastly/Viceroy/blob/f62eb94c470f39df1d25cba3ffb028a3289bbf15/lib/src/wiggle_abi/secret_store_impl.rs#L110-L113). The caller can then try again with an appropriately sized buffer.

### DIFF
--- a/lib/src/wiggle_abi.rs
+++ b/lib/src/wiggle_abi.rs
@@ -158,6 +158,9 @@ impl UserErrorConversion for Session {
                     event!(Level::DEBUG, "Hostcall yielded an error: {}", err);
                 }
             },
+            Error::BufferLengthError { .. } => {
+                event!(Level::DEBUG, "Hostcall yielded an error: {}", e)
+            }
             _ => event!(Level::ERROR, "Hostcall yielded an error: {}", e),
         }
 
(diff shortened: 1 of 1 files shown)


joeshaw:
My 2¢: I think it would be better to lower all the "hostcall yielded" error messages to DEBUG, as you'll see them for various other hostcalls when, for example, running the Go SDK's integration tests.

The fact that a hostcall returned an error to an SDK isn't indicative of a bug in the runtime, or the hostcall.  In the case with `BufferLengthError` it's a negotiation between the runtime and the SDK of the appropriate buffer size, an implementation detail.  But other times it may be caused by unimplemented behavior in Viceroy.  Perhaps most commonly it's caused by user error, passing in bad data.  Those errors should be passed through and reflected in errors by the SDK in the guest code.

draco2003:
+1 for this fix. 

Getting a similar error when code references the [FetchEvent.client.tlsProtocol ](https://js-compute-reference-docs.edgecompute.app/docs/globals/FetchEvent/#instance-properties)and run locally 
image

JakeChampion (approved):

joeshaw (approved):
👍

PR #315, opened 2023-09-27
Return a ValueAbsent for all the downstream-tls related functions instead of a NotAvailable

JakeChampion:
> [!NOTE]  
> This is blocked, waiting on a new release of the fastly crate

Currently these functions return a NotAvailable error, I believe this is because we don't support https within Viceroy.

We do support http though, and http is also supported on Fastly, this patch adds an implementation for the downstream-tls functions which is valid for http (non-tls) connections. As Viceroy does not yet support https, we _could_ say this is a full implementation as-of-today.

In the future if/when Viceroy supports https, we would need to come back to these functions and make them return the correct, expected values based on the incoming connection (tls or not), which is why I have left "FIXME" comments on all of these.

### DIFF
--- a/lib/src/wiggle_abi/req_impl.rs
+++ b/lib/src/wiggle_abi/req_impl.rs
@@ -143,14 +143,14 @@ impl FastlyHttpReq for Session {
         Err(Error::NotAvailable("Client original header fingerprint"))
     }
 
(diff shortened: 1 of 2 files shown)


awortman-fastly (approved):
ahh great, thank you for including asserts in `downstream-req.rs` :pray:

PR #318, opened 2023-10-06
Revamp profile handling CLI flags

fgsch:
This PR moves profiling to `--profile`, and replaces `--profile-guest=[path]` in favour of `--profile=guest[,path]`, unifying the profiling options (and matching Wasmtime). Most of the code is adapted from Wasmtime.

While here I've removed some unncessary lifetimes and polished the code, mostly courtesy of clippy (the 2 last commits).

The PR is related to and was built on top of https://github.com/fastly/Viceroy/pull/316. It might make sense to review that one first.

### DIFF
--- a/cli/src/opts.rs
+++ b/cli/src/opts.rs
@@ -48,11 +48,6 @@ pub struct ServeArgs {
     #[arg(long = "addr")]
     socket_addr: Option<SocketAddr>,
 
-    /// Whether to profile the wasm guest. Takes an optional directory to save
-    /// per-request profiles to
-    #[arg(long, default_missing_value = "guest-profiles", num_args=0..=1, require_equals=true)]
-    profile_guest: Option<PathBuf>,
-
     #[command(flatten)]
     shared: SharedArgs,
 }
@@ -61,11 +56,6 @@ pub struct ServeArgs {
 pub struct RunArgs {
     #[command(flatten)]
     shared: SharedArgs,
-
-    /// Whether to profile the wasm guest. Takes an optional filename to save
-    /// the profile to
-    #[arg(long, default_missing_value = "guest-profile.json", num_args=0..=1, require_equals=true)]
-    profile_guest: Option<PathBuf>,
 
     /// Args to pass along to the binary being executed.
     #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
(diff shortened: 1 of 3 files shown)


lib/src/wiggle_abi/req_impl.rs
```diff
@@ -95,9 +95,9 @@ impl FastlyHttpReq for Session {
     }
 
     #[allow(unused_variables)] // FIXME JDC 2023-06-18: Remove this directive once implemented.
-    fn downstream_client_h2_fingerprint<'a>(
```
JakeChampion, line 98: The changes in this file don't look to be related to the description of this pull-request - could we have these in a different pull-request please? 🤞

cli/src/opts.rs
```diff
@@ -48,11 +48,6 @@ pub struct ServeArgs {
     #[arg(long = "addr")]
     socket_addr: Option<SocketAddr>,
 
-    /// Whether to profile the wasm guest. Takes an optional directory to save
-    /// per-request profiles to
-    #[arg(long, default_missing_value = "guest-profiles", num_args=0..=1, require_equals=true)]
-    profile_guest: Option<PathBuf>,
-
```
JakeChampion, line 55: Note: This would be a breaking change as we are removing a flag, we'd need to version accordingly

lib/src/execute.rs
```diff
@@ -84,7 +84,7 @@ impl ExecuteCtx {
         module_path: impl AsRef<Path>,
         profiling_strategy: ProfilingStrategy,
         wasi_modules: HashSet<ExperimentalModule>,
-        guest_profile_path: Option<PathBuf>,
+        guest_profile_path: Option<String>,
```
JakeChampion, line 87: I think we can avoid this breaking change in the viceroy_lib crate by applying this change instead:
```diff
diff --git a/cli/src/main.rs b/cli/src/main.rs
index bcf7524..db46cdf 100644
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -240,7 +240,7 @@ async fn create_execution_context(
         input,
         args.profiling_strategy(),
         args.wasi_modules(),
-        guest_profile_path,
+        guest_profile_path.map(|p| p.into()),
         args.unknown_import_behavior(),
     )?
     .with_log_stderr(args.log_stderr())
diff --git a/lib/src/execute.rs b/lib/src/execute.rs
index bb45d86..df731fe 100644
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -84,7 +84,7 @@ impl ExecuteCtx {
         module_path: impl AsRef<Path>,
         profiling_strategy: ProfilingStrategy,
         wasi_modules: HashSet<ExperimentalModule>,
-        guest_profile_path: Option<String>,
+        guest_profile_path: Option<PathBuf>,
         unknown_import_behavior: UnknownImportBehavior,
     ) -> Result<Self, Error> {
         let config = &configure_wasmtime(profiling_strategy);
@@ -112,7 +112,6 @@ impl ExecuteCtx {
                 engine_clone.increment_epoch();
             }
         })));
-        let guest_profile_path = guest_profile_path.map(PathBuf::from);
 
         Ok(Self {
             engine,
```

fgsch: I've updated this PR to deal with the breaking change.

JakeChampion (approved):

PR #322, opened 2023-10-09
Match the number of memories to the number of core instances

### LINKED ISSUE #321: 0.9.0 Breaks concurrency
Hello! A number of our engineers use viceroy through the fastly-cli for local development. This morning, a number of them were upgraded to v0.9.0 and immediately had their environments stop working. It seems that any request that is made concurrently with another ongoing request is getting the following error:
```
thread ‘tokio-runtime-worker’ panicked at lib/src/execute.rs:282:27:
failed to run guest: Error instantiating WebAssembly: maximum concurrent memory limit of 1 reached
```
I seem to recall it being 100 at some point in time, when we had to throttle certain types of requests, if that helps. I suspect it's a configurable value, but have no idea where it is configured.

I have confirmed that if I use viceroy v0.8.1 this does not happen.

elliottt:
Fix a bug introduced with the upgrade to wasmtime-13.0.0 where the maximum number of memories was set to `1`. This should have matched the number of core instances we allow, as we permit one memory per core instance.

Fixes #321

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -520,7 +520,7 @@ fn configure_wasmtime(profiling_strategy: ProfilingStrategy) -> wasmtime::Config
     pooling_allocation_config.max_core_instance_size(MB);
 
     // Core wasm programs have 1 memory
-    pooling_allocation_config.total_memories(1);
+    pooling_allocation_config.total_memories(100);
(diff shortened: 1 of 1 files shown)


JakeChampion (approved):
Later we should add a test to cover this regression if possible

PR #327, opened 2023-10-23
🚚 Support for trailers and trailer modification

acw:
This branch adds support for trailers and trailer modification through Viceroy, all of which should be considered experimental for the time being, as we explore the right interfaces for this work.

I've validated this version against grpcurl and an open grpc interface.


lib/compute-at-edge-abi/typenames.witx
```diff
@@ -55,7 +55,14 @@
...
         ;;; resources permitted. For example, creating too many response handles.
-        $limitexceeded))
+        $limitexceeded
+        ;;; Resource temporarily unavailable
+        ;;;
+        ;;; This is returned when an attempting to retrieve a resource that is not yet available.
+        ;;; For example when attempting to read trailers from a Body that has not yet been consumed.
+        $again
+    )
+)
```
acfoltzer, line 65: Match formatting of internal copy
```suggestion
        $limitexceeded
        ;;; Resource temporarily unavailable
        ;;;
        ;;; This is returned when an attempting to retrieve a resource that is not yet available.
        ;;; For example when attempting to read trailers from a Body that has not yet been consumed.
        $again))
```

lib/src/body.rs
```diff
@@ -190,7 +194,25 @@ impl HttpBody for Body {
                             // popped
                             //
                             // TODO ACF 2020-06-01: do something with the body's trailers at this point
```
acfoltzer, line 196: Always nice to retire a 3-year old TODO

```suggestion
```

lib/src/body.rs
```diff
@@ -266,14 +289,23 @@ impl HttpBody for Body {
             }
         }
 
+        // With no more chunks arriving we can mark trailres as being ready.
```
acfoltzer, line 292:
```suggestion
        // With no more chunks arriving we can mark trailers as being ready.
```

acfoltzer (changes requested):
Some small fixes, then good to go imo 👍🏻

acfoltzer (approved):
That Windows thing is weird. We should be on the lookout for reports of it behaving strangely

PR #328, opened 2023-10-26
DEVLIB-1257: Rename Compute@Edge to Compute

doramatadora:
JIRA ticket: [DEVLIB-1257](https://fastly.atlassian.net/browse/DEVLIB-1257)

Please review and notify DevRel (and @Integralist) of the new Viceroy release 🙌🏼  

[DEVLIB-1257]: https://fastly.atlassian.net/browse/DEVLIB-1257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kpfleming (approved):
Once the style guide has been published we'll have to revisit some of this, but it's not worth changing right now.

README.md
```diff
@@ -1,12 +1,12 @@
 # Viceroy
 
-Viceroy provides local testing for developers working with Compute@Edge. It
-allows you to run services written against the Compute@Edge APIs on your local
+Viceroy provides local testing for developers working with Compute. It
```
ulyssa, line 3: I think that this would read clearer if the first mention was "Fastly Compute" instead.

cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "viceroy"
-description = "Viceroy is a local testing daemon for Compute@Edge."
+description = "Viceroy is a local testing daemon for Compute."
```
ulyssa, line 3: I think the package description should probably also say Fastly Compute.

README.md
doramatadora, line 3:
```suggestion
Viceroy provides local testing for developers working with Fastly Compute. It
```

lib/compute-at-edge-abi/README.md
```diff
@@ -1,6 +1,6 @@
 # 🔗 compute-at-edge-abi
```
JakeChampion, line 1: @dora and @kpfleming - do we want to rename this and the folder?

kpfleming: Oh... good question. That's going to be a breaking change for anything which consumes the content here?

kpfleming: Maybe not.. it appears that the WITX files are incorporated into the crate but the directory name is not part of the crate.

kpfleming: But one of the WITX files is named `compute-at-edge.witx`.

PR #332, opened 2023-11-10
Add delete_async hostcall for kv

computermouth:
Tested good on the local machine with an sdk which also has the hostcall


lib/src/session/async_item.rs
computermouth, line 1: Not 100% sure about this change. I needed to differentiate the 
```

aturon: Yep, this is exactly the right thing to do. Not the prettiest, but things can get a bit messy when you go far enough with the type system :)

aturon (approved):
This looks great! ✨

Issue #336, opened 2023-12-12
Support for arbitrary (or at least missing) ecp environment variables


katef:
#146 requested exposing the host machine's environment within the wasm runtime, and was rejected because that behaviour would not match the production network. Quoting @cratelyn there:
> keeping the environment in which a Wasm program runs in parity with production is an important goal for Viceroy

We're missing a few environment variables that we have in prod on ecp proper:
https://developer.fastly.com/reference/compute/ecp-env/

We could hardcode these to dummy values:
```diff
@@ -115,10 +115,19 @@ pub(crate) fn create_store(
 fn make_wasi_ctx(ctx: &ExecuteCtx, session: &Session) -> Result<WasiCtx, anyhow::Error> {
     let mut wasi_ctx = WasiCtxBuilder::new();
 
-    // Viceroy provides a subset of the `FASTLY_*` environment variables that the production
+    // Viceroy provides the same `FASTLY_*` environment variables that the production
     // Compute platform provides:
 
     wasi_ctx
+        // These variables are stubbed out for compatibility
+        .env("FASTLY_CACHE_GENERATION", "0")?
+        .env("FASTLY_CUSTOMER_ID", "0000000000000000000000")?
+        .env("FASTLY_POP", "XXX")?
+        .env("FASTLY_REGION", "Somewhere")?
+        .env("FASTLY_SERVICE_ID", "0000000000000000000000")?
+        .env("FASTLY_SERVICE_VERSION", "0")?
+        .env("FASTLY_TRACE_ID", "00000000-0000-0000-0000-000000000000")?
+
```

But I think adding support for an array of environment variables in the config syntax might make sense, which gives the ability to set these explicitly. I think that's in the spirit of parity with prod, without exposing the entire host machine's environment. I don't know how to do this, or if it's actually a good idea.

fgsch:
I think as a first step we should provide some defaults as you suggested above.
We can have a separate discussion whether we want to or makes sense to expose this in the configuration.

katef:
@fgsch I agree, I implemented the stubbed-out values in #337. I'll leave this ticket open for the config file overrides.

PR #337, opened 2023-12-13
Add stubbed-out values for remaining `FASTLY_` environment variables.

katef:
We already had a few here, with values that make sense. I expect these ones were missing because their values don't make sense, I'm just hardcoding something reasonable. What matters for me is that they're present at all, and to match production ecp in what's present.

My intention is that a user would then be able to override these stub values by explicitly setting things from the config file (see https://github.com/fastly/Viceroy/issues/336 about that), but that there's no way to ever have these not set. Because that matches production.

### DIFF
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -115,10 +115,17 @@ pub(crate) fn create_store(
 fn make_wasi_ctx(ctx: &ExecuteCtx, session: &Session) -> Result<WasiCtx, anyhow::Error> {
     let mut wasi_ctx = WasiCtxBuilder::new();
 
-    // Viceroy provides a subset of the `FASTLY_*` environment variables that the production
+    // Viceroy provides the same `FASTLY_*` environment variables that the production
     // Compute platform provides:
 
     wasi_ctx
+        // These variables are stubbed out for compatibility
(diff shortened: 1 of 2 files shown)


lib/src/linking.rs
```diff
@@ -115,10 +115,17 @@ pub(crate) fn create_store(
 fn make_wasi_ctx(ctx: &ExecuteCtx, session: &Session) -> Result<WasiCtx, anyhow::Error> {
     let mut wasi_ctx = WasiCtxBuilder::new();
 
-    // Viceroy provides a subset of the `FASTLY_*` environment variables that the production
+    // Viceroy provides the same `FASTLY_*` environment variables that the production
```
elliottt, line 118: Would you mind clarifying here that values will be defaulted to stub values when they can't be inherited?

elliottt (approved):
Looks great to me, with a modification to the comment about the env vars. Thanks!

lib/src/linking.rs
katef: Oh, but I don't think they are inherited, are they? if you mean from the host machine's environment

elliottt: You're totally right -- I misread this as being lookups in the env. Ignore that original request :sweat_smile:

PR #341, opened 2023-12-19
return an empty result when geolocation data is not found

### LINKED ISSUE #339: Geo lookup returns FastlyStatus::None when not found, which is different from Compute
When calling the `geo::lookup` hostcall, if there is no match for the provided IP, Viceroy [returns `FastlyStatus::None`](https://github.com/fastly/Viceroy/blob/8f2ea50bff09edead16aeb16b14d6cb731bdf9b7/lib/src/wiggle_abi/geo_impl.rs#L30).  Fastly Compute, however, returns no error and an empty buffer.  Viceroy should match this behavior.

joeshaw:
This matches the Fastly Compute behavior.

As a result, we don't need the `GeolocationError` type anymore and this
commit also removes it.

Fixes #339.

### DIFF
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -93,9 +93,6 @@ pub enum Error {
     DeviceDetectionError(#[from] crate::wiggle_abi::DeviceDetectionError),
 
     #[error(transparent)]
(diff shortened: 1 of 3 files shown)


JakeChampion (approved):
LGTM 👍

PR #344, opened 2024-02-19
Sync hostcalls in preparation for new SDK release

ulyssa:
This PR syncs the `.witx` files with the hostcalls that are currently supported in Compute, and which will be included in the next Rust SDK release. The two new hostcalls are `downstream_tls_ja4`, which I've stubbed out to be similar to `downstream_tls_ja3_hash`, and `known_length`, which I've added a basic implementation for.


acfoltzer (approved):
Nice, thank you!

PR #349, opened 2024-03-12
Consistently use Error::NotAvailable instead of Unsupported

joeshaw:
This PR switches most uses of `Error::Unsupported` to `Error::NotAvailable`, and translates both to `FastlyStatus::Unsupported`.

`Unsupported` is meant to express things that are not supported by the platform, for example trying to make a request with "HTTP/4".

`NotAvailable` is meant to express things that are not available or unimplemented in Viceroy.

Previously, `Unsupported` would return `FastlyStatus::Unsupported` and `NotAvailable` would return the generic `FastlyStatus::Error`.  This led to a bad user experience for the `NotAvailable` case, as the user would see an opaque and generic "Error" message rather than indicating an operation was "Unsupported".

This PR (1) changes `Error::NotAvailable` to map to `FastlyStatus::Unsupported` and (2) changes all our uses of `Error::Unsupported` for things that are not implemented in Viceroy to be `Error::NotAvailable`.

### DIFF
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -159,7 +159,7 @@ impl Error {
             Error::BufferLengthError { .. } => FastlyStatus::Buflen,
             Error::InvalidArgument => FastlyStatus::Inval,
             Error::ValueAbsent => FastlyStatus::None,
-            Error::Unsupported { .. } => FastlyStatus::Unsupported,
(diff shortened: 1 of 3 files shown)


cee-dub (approved):
LGTM (though I'm pretty new to Rust)

JakeChampion (approved):

PR #351, opened 2024-03-18
draft: generate certificate and serve over http and https

JakeChampion:
pushing this up to see if there is interest in seeing this feature land and if this approach is along the right tracks

TODO:
- [ ] Allow certificate to be provided via cli flag
- [ ] Allow certificate to be provided via fastly.toml
- [ ] Allow http and https port to be defiend via cli flag

### DIFF
--- a/lib/src/service.rs
+++ b/lib/src/service.rs
@@ -1,7 +1,9 @@
 //! Service types.
 
+use rcgen::generate_simple_self_signed;
 use {
     crate::{body::Body, execute::ExecuteCtx, Error},
+    axum_server::tls_rustls::RustlsConfig,
     futures::future::{self, Ready},
     hyper::{
         http::{Request, Response},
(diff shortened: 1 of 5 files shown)


kpfleming:
This seems really promising! In addition to the TODOs above, have you considered allowing the user to provide a list of SAN(s) to be included in the cert, so that they can test with their 'real' name for the service?

lib/src/service.rs
```diff
@@ -62,12 +64,41 @@ impl ViceroyService {
...
+        let cert = generate_simple_self_signed(subject_alt_names).unwrap();
+
+        let config = RustlsConfig::from_pem(
+            cert.serialize_pem().unwrap().into_bytes(),
+            cert.serialize_private_key_pem().into_bytes(),
+        )
+        .await
+        .unwrap();
+
+        let addr = SocketAddr::new(IpAddr::V4(std::net::Ipv4Addr::new(127, 0, 0, 1)), 7879);
```
kpfleming, line 95: Is there a benefit to this more-verbose construct, compared to the simpler one on line 78?

PR #354, opened 2024-04-02
Local integration tests

dpetrick:
# Summary
- Adds the ability to read Endpoint messages programmatically by having them write to an in-memory buffer that can be read from outside of the execution context after an invocation. Technically this could have been intercepted from stdout before, but this is much more convenient.
- Adds the ability to handle requests in-memory, instead of reaching out to an origin via HTTP. This is done by adding an in-memory handler to backends to which Requests will be passed.
- Adds the ability to intercept dynamic backend registrations and traffic, which in combination with in-memory backend handlers now allows for in-memory handling of dynamic backend requests.
- Cache API support. Allows keeping states in between invocations. No transaction support, plus some calls unused by stellate are not yet implemented. Very simplistic cache implementation that is not really suited for long-running processes, but perfectly
- Adds the ability to wait for guest code to complete, which is necessary to surface execution errors after an initial response has been sent to client (especially true if stream processing is done).

# Implementation notes
- We could implement service chains with this harness. However, our focus is on one-layer.
- In-memory backends are super basic and don't perform any checks. It would be worth expanding upon the concept.
- Endpoint listeners are also super basic in concept and execution, they probably require a second look as well.
- Caches use a lot of lock unwraps.


dpetrick:
Note that we'll record a video on the ins and outs / requirements and we'll also reach out to you directly via Slack to start a discussion. This is not expected to get merged, merely showcase.

dpetrick:
Closing in favor of https://github.com/fastly/Viceroy/pull/589

Issue #360, opened 2024-04-15
Update Wasmtime


acme:
Viceroy currently [uses Wasmtime version 13.0.0](https://github.com/fastly/Viceroy/blob/main/Cargo.toml#L42). 

Wasmtime has [released a number of versions](https://github.com/bytecodealliance/wasmtime/releases) since then (currently 19.0.2) and it would be great to incorporate these improvements.

kpfleming:
Addressed by #366.

Issue #357, opened 2024-04-09
CLI support for an optional ,interval for the guest profiler, to match wasmtime


katef:
In #301 support for guest profiling was added. This corresponds to an equivalent CLI flag in wasmtime introduced in https://github.com/bytecodealliance/wasmtime/pull/6362 (and documented in https://github.com/bytecodealliance/wasmtime/pull/6393).

We have no way to control the profiling interval for Viceroy:

Viceroy: `--profile=guest[,path]`
Wasmtime: `--profile=guest[,path[,interval]]`

This ticket is to add the optional `,interval` part, just like wasmtime has.

posborne:
Implemented with https://github.com/fastly/Viceroy/pull/456

PR #361, opened 2024-04-16
Update wasmtime* crates to 19.0.2

fgsch:
I wanted to have an idea of how much effort, at least superficially, was required to update Viceroy.
The tests are passing but beyond that this is mostly mechanical and untested.

Addresses #360 and I believe is required for #357.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,17 +125,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
-dependencies = [
- "hermit-abi 0.1.19",
- "libc",
- "winapi",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.2.0"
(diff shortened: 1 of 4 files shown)


jameysharp (approved):
I can confirm that the guest-profiler change here is valid: there's an opportunity to collect more data and produce even more useful profiles, but as written this PR preserves the existing behavior, which is good enough.

I'm less sure about switching from `wasmtime-wasi` to `wasi-common`. My impression is that `wasi-common` is becoming deprecated, but I'm not certain of that.

jameysharp:
Oh, and if we want to note anything about user-visible changes in Viceroy release notes, upgrading to Wasmtime 18 or newer gets us name demangling in guest-profiler output, which is a nice quality of life improvement when profiling Rust services especially.

fgsch:
>[..]
> I'm less sure about switching from `wasmtime-wasi` to `wasi-common`. My impression is that `wasi-common` is becoming deprecated, but I'm not certain of that.

Thanks for having a look. I will check this and update the PR accordingly.

elliottt:
Do you mind if we close this in favor of #366? I've made the `wasmtime-wasi` switch in that PR as well.

fgsch:
Closed in favour of #366

PR #365, opened 2024-05-03
Allow specifying top-level `file` and `format` for KVStore

### LINKED ISSUE #364: Allow JSON files as input for Secrets Stores and KVStores
Currently the `fastly.toml` [specification](https://www.fastly.com/documentation/reference/compute/fastly-toml/) allows you to pass in a JSON file for config stores (and earlier, also for dictionaries)

Something like that is extremely handy to be able to do, to e.g. share configuration across multiple fastly.tomls for environments, autogenerating these values from other scripts, not having to check them into CI etc.

Is that something you might be interested in adding back?

GeeWee:
It's very practical to be able to keep all your local configuration in separate files, if they are e.g. generated by automatic scripts. This extends the `fastly.toml` `kv_store` configuration to accept either the current inline API or a new `file` and `format` table, that allows loading of an external JSON file.

Part of #364

### DIFF
--- a/cli/tests/integration/kv_store.rs
+++ b/cli/tests/integration/kv_store.rs
@@ -61,6 +61,36 @@ viceroy_test!(object_stores_backward_compat, |is_component| {
     Ok(())
 });
 
+viceroy_test!(kv_store_allows_fetching_of_key_from_file, |is_component| {
+    // This test checks that we can provide a "format" and a "file"
+    // with a JSON dictionary inside it for an object store (aka KV Store)
+    // and have the KV Store populated with it.git
+    const FASTLY_TOML: &str = r#"
+        name = "kv-store-test"
+        description = "kv store test"
+        authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
+        language = "rust"
+        [local_server]
+        object_stores.empty_store = []
+        object_stores.store_one = { file = "../test-fixtures/data/json-kv_store.json", format = "json" }
+    "#;
+
+    let resp = Test::using_fixture("kv_store.wasm")
+        .adapt_component(is_component)
+        .using_fastly_toml(FASTLY_TOML)?
+        .against_empty()
+        .await?;
+
+    assert_eq!(resp.status(), StatusCode::OK);
+    assert!(to_bytes(resp.into_body())
+        .await
+        .expect("can read body")
+        .to_vec()
+        .is_empty());
+
+    Ok(())
+});
+
 viceroy_test!(object_store_backward_compat, |is_component| {
     // Previously the "object_stores" key was named "object_store" and
     // the "file" key was named "path".  This test ensures that both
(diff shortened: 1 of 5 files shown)


lib/src/config/object_store.rs
```diff
@@ -111,6 +169,34 @@ impl TryFrom<Table> for ObjectStoreConfig {
                     .expect("Lock was not poisoned");
             }
         }
+
         Ok(ObjectStoreConfig(obj_store))
     }
 }
+
+fn read_json_contents(file: &Path) -> Result<HashMap<String, String>, ObjectStoreConfigError> {
```
GeeWee, line 177: This is mostly just copy-pasted from the dictionary implementation with different errors

lib/src/error.rs
```diff
@@ -644,6 +644,16 @@ pub enum ObjectStoreConfigError {
     ObjectStoreError(#[from] crate::object_store::ObjectStoreError),
     #[error("Invalid `key` value used: {0}.")]
     KeyValidationError(#[from] crate::object_store::KeyValidationError),
+    #[error("'{0}' is not a valid format for the config store. Supported format(s) are: 'json'.")]
```
GeeWee, line 647: These are heavily inspired by the dictionary errors

kpfleming:
Thank you for the contribution! We're actually in the midst of some discussions about how to redesign the way that Viceroy handles 'resources' needed by the service, so we'll have to hold off on reviewing this PR for a few weeks until we've reached some decisions there. The good news is that the idea you've presented here is directly in line with our goals :-)

Please stay tuned, we'll get back to you soon.

GeeWee:
Gotcha - thanks for the feedback :)

kpfleming:
Overall this looks fine, but as you can probably see from the conflicts the term 'object store' shouldn't be used for this feature, it should be 'kv store' (in various forms) everywhere. If you get the PR updated I'll give it a formal review and work towards getting it merged. Thanks!

GeeWee:
@kpfleming the test file is reasonably inconsistent (many tests still refer to object store). However I've fixed my changes on top of the latest `main` and changed to use kv-store instead of object store in my tests.

I've rebased and squashed my changes into one commit. Let me know if you don't care about that (e.g. if you'll end up squashing when merging anyways) and I'll save the effort :)

JakeChampion:
Another approach we can take is to point a KV Store to a folder and use the file-paths as the KV Store Keys and the file contents as the value -- This would allow for simpler setup for the scenario where applications are wanting to store their website static assets in a KV Store

JakeChampion:
> Another approach we can take is to point a KV Store to a folder and use the file-paths as the KV Store Keys and the file contents as the value -- This would allow for simpler setup for the scenario where applications are wanting to store their website static assets in a KV Store

I have a branch which implements this at https://github.com/fastly/Viceroy/tree/jake/kv-store-folder but I stalled out on the work unfortunately

GeeWee:
Definitely also an option - I do wonder if keeping parity with the dictionary interface isn't a good thing though, considering 
that we can't be the only migrating I imagine :)

lib/src/config/object_store.rs
```diff
@@ -15,12 +18,69 @@ impl TryFrom<Table> for ObjectStoreConfig {
...
+            // We try to parse either one of them to the same Vec<toml::Value>
+            // to allow them to run through the same validation path further down
+            let file_path = items
+                .as_table()
+                .and_then(|table| table.get("file"))
+                .and_then(|file| file.as_str());
+            let file_format = items
+                .as_table()
+                .and_then(|table| table.get("format"))
+                .and_then(|file| file.as_str());
```
kpfleming, line 32:
```suggestion
                .and_then(|format| format.as_str());
```

kpfleming (changes requested):
A few changes you can commit directly, and some others you can make in a followup commit. You can squash or not squash, whichever is easier for you.

kpfleming:
> > Another approach we can take is to point a KV Store to a folder and use the file-paths as the KV Store Keys and the file contents as the value -- This would allow for simpler setup for the scenario where applications are wanting to store their website static assets in a KV Store
> 
> I have a branch which implements this at https://github.com/fastly/Viceroy/tree/jake/kv-store-folder but I stalled out on the work unfortunately

We've got another PoC implementation around which does the same thing as a WASM component :-)

For now I'm OK with merging the approach in this PR because it doesn't add a new 'style' of data population, it just replicates the existing style offered for config stores. We can talk about offering other styles in the larger topic of redesigning the package manifest/configuration file.

cli/tests/integration/kv_store.rs
```diff
@@ -232,6 +262,80 @@ viceroy_test!(kv_store_bad_configs, |is_component| {
...
+        description = "kv store test"
+        authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
+        language = "rust"
+        [local_server]
+        object_stores.empty_store = []
+        object_stores.store_one = { file = "../test-fixtures/data/ABSOLUTELY_NOT_A_REAL_PATH", format = "json" }
+    "#;
+    match Test::using_fixture("kv_store.wasm").using_fastly_toml(BAD_13_FASTLY_TOML) {
+        Err(e) => assert_eq!(
+            "invalid configuration for 'store_one': No such file or directory (os error 2)",
```
kpfleming, line 318: This isn't the error message produced on Windows, so this test fails when run there. It's good that you added a test to provide more coverage than the code you started with, but this one will be a little tricky to implement :-)

cli/tests/integration/kv_store.rs
```diff
@@ -61,6 +61,36 @@ viceroy_test!(object_stores_backward_compat, |is_component| {
     Ok(())
 });
 
+viceroy_test!(kv_store_allows_fetching_of_key_from_file, |is_component| {
+    // This test checks that we can provide a "format" and a "file"
+    // with a JSON dictionary inside it for an object store (aka KV Store)
+    // and have the KV Store populated with it.git
```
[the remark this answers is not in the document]
GeeWee: Woops!

cli/tests/integration/kv_store.rs
GeeWee: Aah, stupid platform-specific error messages ☹️ 
I've changed the test to simply assert on the first part of the error message as a "good-enough" kind-of thing. I considered just asserting for either of the messages, but I figured that'll be brittle whenever we add another platform.

cli/tests/integration/kv_store.rs
```diff
@@ -61,6 +61,36 @@ viceroy_test!(object_stores_backward_compat, |is_component| {
     Ok(())
 });
 
+viceroy_test!(kv_store_allows_fetching_of_key_from_file, |is_component| {
+    // This test checks that we can provide a "format" and a "file"
+    // with a JSON dictionary inside it for an object store (aka KV Store)
```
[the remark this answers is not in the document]
GeeWee: Gotcha. I've applied your suggestions and also changed this block here to only refer to KVStore.

cli/tests/integration/kv_store.rs
```diff
@@ -61,6 +61,36 @@ viceroy_test!(object_stores_backward_compat, |is_component| {
...
+    // and have the KV Store populated with it.git
+    const FASTLY_TOML: &str = r#"
+        name = "kv-store-test"
+        description = "kv store test"
+        authors = ["Jill Bryson <jbryson@fastly.com>", "Rose McDowall <rmcdowall@fastly.com>"]
```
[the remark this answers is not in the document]
GeeWee: Done!

GeeWee:
@kpfleming I believe I've incorporated all of your feedback :)

kpfleming (dismissed):

kpfleming:
The tests all pass now, I'll add an entry for CHANGELOG.md and get this merged.

kpfleming (approved):

PR #366, opened 2024-05-07
Upgrade to wasmtime-20, and switch to wasmtime-wasi

elliottt:
Update the wasmtime dependency to wasmtime-20, and also remove the dependency on wasi-common in favor of using the preview1-on-preview2 implementation provided by wasmtime-wasi.

There was also a change in wiggle's borrow checker for the 20.0.0 release that affected our host call implementations directly: borrows are no longer to a region, but the whole guest memory. This caused a bit of churn in the wiggle_abi tree, and I'm not completely certain that all issues have been resolved, as the test suite can only test so many paths through the host calls. For reference, here's the wiggle refactoring PR in wasmtime: https://github.com/bytecodealliance/wasmtime/pull/8277.

### DIFF
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -5,11 +5,9 @@ use {
         config::ExperimentalModule, execute::ExecuteCtx, logging::LogEndpoint, session::Session,
         wiggle_abi, Error,
     },
-    anyhow::Context,
     std::collections::HashSet,
-    wasi_common::{pipe::WritePipe, WasiCtx},
     wasmtime::{GuestProfiler, Linker, Store, UpdateDeadline},
-    wasmtime_wasi::WasiCtxBuilder,
+    wasmtime_wasi::{preview1::WasiP1Ctx, WasiCtxBuilder},
     wasmtime_wasi_nn::WasiNnCtx,
 };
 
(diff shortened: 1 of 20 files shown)


lib/src/logging.rs
```diff
@@ -57,7 +57,10 @@ impl LogEndpoint {
         }
         to_write.push(b'\n');
 
-        LOG_WRITER.lock().unwrap().write_all(&to_write)
+        eprintln!("locking");
```
athomason, line 60: Were these prints meant to be in the final product?

elliottt: Nope, thank you for catching those!

lib/src/linking.rs
```diff
@@ -102,7 +105,7 @@ pub(crate) fn create_store(
     store.set_epoch_deadline(1);
     store.epoch_deadline_callback(|mut store| {
         if let Some(mut prof) = store.data_mut().guest_profiler.take() {
-            prof.sample(&store);
+            prof.sample(&store, std::time::Duration::ZERO);
```
acw, line 108: I'm just going to highlight this, because a sample duration of zero kind of confuses me. I guess it can't be the epoch time because we may have taken other samples during that epoch? Is there a reason ("because it's too slow, Adam") to not keep a time-since-last-sample, or something?

acw (approved):
One small question, which may be just fine, but otherwise looks good to me.

lib/src/linking.rs
elliottt: This delta parameter is the amount of cpu time that was taken since the last `sample` call. Passing zero is called out as okay in the docs for cases where cpu usage information isn't needed. We can always start tracking that for more precise profiles in the future, but this seemed like a reasonable first step :)

PR #368, opened 2024-05-20
Fix header_values_set handling in the component implementation

elliottt:
Make the handling of header values given to the `header-values-set` functions in the component implementation consistent with the existing wiggle implementation. Additionally, use `next_back` instead of reversing and skipping the first element.


acw (approved):
Took me a bit to follow the inlining, but makes sense to me.

PR #369, opened 2024-05-20
Update to wasmtime-21

elliottt:
Update to wasmtime 21.0.0.

The most substantial changes were to the use of the `component::bindgen!` macro, which now no longer automatically generates host bindings that are allowed to trap. Additionally, the generated `add_to_linker` functions are no longer present, and we instead need to use `add_to_linker_get_host`, which requires a new projection function to be supplied.


acw (approved):
LGTM!

PR #370, opened 2024-05-20
Remove the trappable_error_type

elliottt:
The `trappable_error_type` argument to `wasmtime::bindgen!` became optional in wasmtime-21, and given that we never expect to trap in any of our host implementations, let's remove that additional complexity.

The meat of this change is removing the custom `FastlyError` type, and replacing it with `types::Error` which is generated from compute.wit. This simplifies some functions like the `v2` suffixed request functions, but in most cases ended up being a renaming from `FastlyError` to `types::Error`.

### DIFF
--- a/lib/src/component/http_req.rs
+++ b/lib/src/component/http_req.rs
@@ -2,7 +2,6 @@ use {
     super::{
         fastly::api::{http_req, http_types, types},
         headers::write_values,
-        FastlyError,
     },
     crate::{
         config::{Backend, ClientCertInfo},
@@ -54,7 +53,7 @@ impl http_req::Host for Session {
         &mut self,
         h: http_types::RequestHandle,
         max_len: u64,
-    ) -> Result<String, FastlyError> {
+    ) -> Result<String, types::Error> {
         let req = self.request_parts(h.into())?;
         let req_method = &req.method;
 
@@ -73,7 +72,7 @@ impl http_req::Host for Session {
         &mut self,
         h: http_types::RequestHandle,
         max_len: u64,
-    ) -> Result<String, FastlyError> {
+    ) -> Result<String, types::Error> {
         let req = self.request_parts(h.into())?;
         let req_uri = &req.uri;
         let res = req_uri.to_string();
@@ -95,7 +94,7 @@ impl http_req::Host for Session {
         _tag: http_req::CacheOverrideTag,
         _ttl: u32,
         _stale_while_revalidate: u32,
-    ) -> Result<(), FastlyError> {
+    ) -> Result<(), types::Error> {
         // For now, we ignore caching directives because we never cache anything
         Ok(())
     }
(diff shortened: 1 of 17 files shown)


lib/src/component/http_req.rs
```diff
@@ -396,32 +398,26 @@ impl http_req::Host for Session {
         h: http_types::RequestHandle,
         b: http_types::BodyHandle,
         backend_name: String,
-    ) -> Result<
-        // This return type is a little surprising. Please see the [error-detail] note at the top
-        // of the file for an explanation.
```
acw, line 401: yeah, glad to see this go

lib/wit/deps/fastly/compute.wit
```diff
@@ -136,10 +136,10 @@ interface uap {
...
 
   resource user-agent {
-    family: func(max-len: u64) -> string;
-    major: func(max-len: u64) -> string;
-    minor: func(max-len: u64) -> string;
-    patch: func(max-len: u64) -> string;
+    family: func(max-len: u64) -> result<string, error>;
+    major: func(max-len: u64) -> result<string, error>;
+    minor: func(max-len: u64) -> result<string, error>;
+    patch: func(max-len: u64) -> result<string, error>;
```
acw, line 142: This bit seems unrelated; is this an intentional inclusion in this PR?

lib/src/component/http_req.rs
```diff
@@ -396,32 +398,26 @@ impl http_req::Host for Session {
...
-    > {
+    ) -> Result<http_types::Response, (Option<http_req::SendErrorDetail>, types::Error)> {
         // This initial implementation ignores the error detail field
-        match self.send(h, b, backend_name).await {
-            Ok(r) => Ok(Ok(r)),
-            Err(e) => e.with_empty_detail(),
-        }
+        self.send(h, b, backend_name)
+            .await
+            .map_err(types::Error::with_empty_detail)
```
acw, line 405: If all errors here are going to return `(None, e)`, why are we including the `None`? Is this a future expansion thing?

acw (changes requested):
Just a couple questions!

lib/wit/deps/fastly/compute.wit
elliottt: Yes. This was an omission from the original pr: those functions pass in a max-length for the resulting string, but had no way to signal an error. Given that I was making these changes, I updated this at the same time.

lib/src/component/http_req.rs
elliottt: Yep. The wiggle implementation always sets the error-detail to a default value, but in the wit version, I opted to represent that as an `option<send-error-detail>` instead. Returning `None` is the same as the wiggle version, but I didn't want to break our ability to return the additional detail later.

acw (approved):
Works for me!

PR #371, opened 2024-05-21
Remove an unnecessary Arc from the Session

elliottt:
`ObjectStores` internally is an `Arc` to an `RwLock`, so the outer `Arc` present in `Session` is not necessary. This PR removes that unnecessary indirection, as `ObjectStore` already derives `Clone`.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -86,7 +86,7 @@ pub struct ExecuteCtx {
     /// The ID to assign the next incoming request
     next_req_id: Arc<AtomicU64>,
     /// The ObjectStore associated with this instance of Viceroy
-    object_store: Arc<ObjectStores>,
+    object_store: ObjectStores,
     /// The secret stores for this execution.
     secret_stores: Arc<SecretStores>,
     // `Arc` for the two fields below because this struct must be `Clone`.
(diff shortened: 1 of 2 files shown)


athomason (approved):
LGTM. Rust idiomatic q: I notice that SecretStores has `HashMap<String, SecretStore>`, is one placement of the Arc preferable to the other?

elliottt:
> LGTM. Rust idiomatic q: I notice that SecretStores has `HashMap<String, SecretStore>`, is one placement of the Arc preferable to the other?

I think it depends on what the desired external use of the value is: putting the `Arc` inside means that the value is clonable, while leaving it out means that the user of the type has more control over data indirection.

acw:
> LGTM. Rust idiomatic q: I notice that SecretStores has `HashMap<String, SecretStore>`, is one placement of the Arc preferable to the other?

I don't know if I've detected a general law, and I don't think clippy has never directed me in one direction or another. I can say that my general rule is:
- If the data in the  object can be freely cloned without any sort of problem, then don't use an Arc.
- If some/all of the data in the object needs to be shared across concurrent threads/tasks and I care about *how* parts of it are synchronized, then I put the lock in the object. So, if I really think the right lock is `RwLock` and always will be `RwLock`.
- I'm increasingly of the opinion that if I don't care exactly how things are locked, but I do think stuff will be shared, then I make sure the object doesn't implement `Clone` and have my callers work it out.

athomason:
Thanks!

PR #372, opened 2024-05-22
Have kv-store#open return an optional handle

elliottt:
Modify the definition of the `open` function in the `kv-store` interface to return an optional handle instead of wrapping that case into the error.

### DIFF
--- a/lib/src/component/kv_store.rs
+++ b/lib/src/component/kv_store.rs
@@ -2,7 +2,6 @@ use {
     super::fastly::api::{http_types, kv_store, types},
     crate::{
         body::Body,
-        error,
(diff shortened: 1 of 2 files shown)


guybedford (commented):
There are various functions where the `optional-none` error is used. Perhaps this error code can be entirely deprecated in the WIT.

elliottt:
> There are various functions where the `optional-none` error is used. Perhaps this error code can be entirely deprecated in the WIT.

Yep, I'd like to deprecate that error value, but haven't audited all the calls yet to make sure they're all migrated.

aturon (approved):
Yay, it's so exciting to start moving to higher-level types 🤩

PR #373, opened 2024-05-23
Add stubs for new "busy handle" APIs for the core cache

aturon:
As with existing caching APIs, these are just stubs for now, needed to support new SDK releases.

### DIFF
--- a/lib/wit/deps/fastly/compute.wit
+++ b/lib/wit/deps/fastly/compute.wit
@@ -1008,7 +1008,7 @@ interface cache {
 
   /// Close an ongoing interaction with the cache.
   ///
-  /// If the cache handle state includes the `$must_insert_or_update` (and hence no insert or
+  /// If the cache handle state includes the `must-insert-or-update` (and hence no insert or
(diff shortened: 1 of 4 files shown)


lib/wit/deps/fastly/compute.wit
```diff
@@ -985,6 +1003,14 @@ interface cache {
   /// Useful if there is an error before streaming is possible, e.g. if a backend is unreachable.
   transaction-cancel: func(handle: handle) -> result<_, error>;
 
+  /// Close an interaction with the cache that has not yet finished request collapsing.
+  close-busy: func(handle: busy-handle) -> result<_, error>;
+
+  /// Close an ongoing interaction with the cache.
+  ///
+  /// If the cache handle state includes the `$must_insert_or_update` (and hence no insert or
```
elliottt, line 1011:
```suggestion
  /// If the cache handle state includes the `must-insert-or-update` (and hence no insert or
```

elliottt (dismissed):
Looks great to me! I made one suggestion to update a comment to use the wit name of the flag instead of the witx one.

PR #377, opened 2024-05-31
Plumb through reading the FASTLY_KEY env variable to compute modules

computermouth:
This is mostly for some internal testing at the moment, but potentially of use to end users who want to utilize fastly services or features which aren't included in the SDK yet. Also allows for local testing against production services.


elliottt:
Looks like tests are failing the formatting step, could you run `cargo fmt` and commit the results?

elliottt (approved):
I think this seems reasonable, and is a good way to start distinguishing between components that target the viceroy environment, vs the production environment.

cceckman-at-fastly:
In the interest of cleaning up the review queue, I'm going to close this PR. If you'd like to address the merge issues & reopen / submit under a new PR please do so!

PR #378, opened 2024-06-05
Return HttpInvalid error instead of FastlyStatus::Error when failed to build Uri

arayaryoma:
[Fastly Rust SDK maps `FastlyStatus::Error` to `BackendCreationError::NameInUse`](https://docs.rs/fastly/0.10.1/src/fastly/backend/builder.rs.html#82). 
By this, when we passed invalid host name value to `fastly::Backend::builder` func, we always received `NameInUse` error.
It's a misleading behavior for developers, who will assume they register the same backend multiple times.

In this PR, I defined a new internal Error `InvalidBackendUrl` and map it as `FastlyStatus::HttpInvalid`.
It will help developers to awake they are using invalid URI to build the dynamic backend.

### before
```rust
let backend = match fastly::Backend::builder("example", "http://example.com").finish() {
    Ok(backend) => backend,
    Err(e) => {
        println!("Error: {:?}", e); // Error: NameInUse
        return Err(e);
    }
};
```

### after
```rust
let backend = match fastly::Backend::builder("example", "http://example.com").finish() {
    Ok(backend) => backend,
    Err(e) => {
        println!("Error: {:?}", e); // Error: HostError(FastlyStatus::HTTP_INVALID_ERROR)
        return Err(e);
    }
};
```

### DIFF
--- a/lib/src/component/error.rs
+++ b/lib/src/component/error.rs
@@ -175,6 +175,7 @@ impl From<error::Error> for types::Error {
             Error::DictionaryError(e) => e.into(),
             Error::ObjectStoreError(e) => e.into(),
             Error::SecretStoreError(e) => e.into(),
+            Error::InvalidBackendUrl => types::Error::HttpInvalid,
             // All other hostcall errors map to a generic `ERROR` value.
             Error::AbiVersionMismatch
             | Error::BackendUrl(_)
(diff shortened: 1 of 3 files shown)


kpfleming:
This is going in a good direction, but the name of the error itself seems incorrect: the parameter to the backend builder is not a URL, so the error cannot indicate that the user provided an 'invalid backend URL' (even though the provided value is used in constructing a URL). Should it be `InvalidBackendSpecification` or something similar?

kpfleming:
@kailan Has #613 made this PR obsolete?

PR #379, opened 2024-06-10
Switch to reading dictionaries during the `fastly_dictionary_open` call

elliottt:
Read dictionaries during the `open` host call instead of once for each `get` call. This improves viceroy performance for large dictionaries, and more closely matches the semantics of the production environment, as changes made while the dictionary is open won't be reflected in subsequent `get` calls.

Additionally, remove the `DictionaryName` newtype, as it wasn't offering anything over using a `String` directly.

co-authored-by: Jamey Sharp <jsharp@fastly.com>

### DIFF
--- a/lib/src/component/config_store.rs
+++ b/lib/src/component/config_store.rs
@@ -16,10 +16,7 @@ impl config_store::Host for Session {
         name: String,
         max_len: u64,
     ) -> Result<Option<String>, types::Error> {
-        let dict = self
-            .dictionary(store.into())?
-            .contents()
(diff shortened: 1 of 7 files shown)


acfoltzer (approved):
Great improvement on the reduced parsing!

I have mixed feelings about removing `DictionaryName`; I suspect it was set up as a newtype in order to potentially encode the platform restrictions on dictionary names. But if it's not being used for that right now, we might as well get rid of it... there's already some deferred work hiding here to differentiate between dictionaries and config stores, and enforcing name restrictions would probably be better done at that point.

PR #381, opened 2024-06-11
Fix buffer-len handling in the component adapter

elliottt:
Fix handling of the buffer-length errors in the component adapter by turning the compute.wit error type into a variant. This allows us to pass out the length that would have allowed the operation to succeed when raising the `Error::BufferLen` error.

This PR also adds the necessary plumbing for running additional tests as adapted components, which should help catch more of these bugs as more of the test suite is enabled for adapted components.

One change was required to the `response` test for it to work with the component model: testing that too-long header names returned an error required that we switch to explicitly constructing those header values, instead of passing an invalid length. This is due to the requirement that we be able to construct a valid string from the buffer/length arguments, and would trap when using the component adapter otherwise.


acfoltzer (approved):
Ah, yeah, it makes sense that this would need to change. The convention of using `nwritten_out` to communicate this value is... very much not Componenty. This looks good

PR #382, opened 2024-06-11
Fix trap handling when running components

elliottt:
The component implementation didn't have support for either the `test-fatalerror-config` feature, or the ability to treat the `Error::FatalError` variant as a trap. This PR fixes that by doing the following:

* Allowing `header_values_get` to trap based on the `test-fatalerror-config` feature
* Introducing the `TrappableError` type for treating `FatalError` as a trap
* Adding a variant of the `trap-test` that uses the component adapter

The remaining difference to the wiggle implementation is that only functions listed in `trappable_errors` will be able to emit traps for `FatalError`, though the only case we currently use that in is for `test-fatalerror-config`, so it seemed reasonable to not switch everything to `TrappableError`.

### DIFF
--- a/lib/src/component/http_resp.rs
+++ b/lib/src/component/http_resp.rs
@@ -1,7 +1,8 @@
 use {
     super::fastly::api::{http_resp, http_types, types},
-    super::headers::write_values,
+    super::{headers::write_values, types::TrappableError},
     crate::{error::Error, session::Session},
+    cfg_if::cfg_if,
     http::{HeaderName, HeaderValue},
     hyper::http::response::Response,
     std::str::FromStr,
@@ -79,6 +80,8 @@ impl http_resp::Host for Session {
         max_len: u64,
         cursor: u32,
     ) -> Result<Option<(Vec<u8>, Option<u32>)>, types::Error> {
+        {}
+
         let headers = &self.response_parts(h.into())?.headers;
 
         let (buf, next) = write_values(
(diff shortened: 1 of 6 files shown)


lib/src/component/http_resp.rs
```diff
@@ -79,6 +80,8 @@ impl http_resp::Host for Session {
         max_len: u64,
         cursor: u32,
     ) -> Result<Option<(Vec<u8>, Option<u32>)>, types::Error> {
+        {}
+
```
acfoltzer, line 84: Uh oh, weird merge artifact maybe?

lib/src/component/mod.rs
```diff
@@ -12,6 +12,14 @@ component::bindgen!({
...
         "wasi:cli": wasmtime_wasi::bindings::cli,
     },
+
+    trappable_error_type: {
+        "fastly:api/types/error" => types::TrappableError,
+    },
+
+    trappable_imports: [
+        "header-values-get"
+    ],
```
acfoltzer, line 22: It's a little funky that these types will be different even in the non-trapping config, but I think this probably the lightest touch possible for making the two modes work out (vs duplicating the macro invocation entirely between both configs... ugh)

lib/src/component/http_resp.rs
acfoltzer:
```suggestion
```

acfoltzer (changes requested):
Looks good except for weird stray braces

lib/src/component/mod.rs
elliottt: It also makes it so that we don't have to deal with nested `Result` values in the return type for `header-values-get`, which is much easier for me to think about :sweat_smile:

acfoltzer (approved):
🪤

PR #383, opened 2024-06-13
add adaptive buffer support for geo + device detection lookups

joeshaw:
Write out the required buffer size to the nwritten_out value if the
buffer is too small to fit the value.


cee-dub (approved):
Thanks!!

joeshaw:
Sorry, should have mentioned that this matches the behavior on the production Compute platform.

PR #384, opened 2024-06-13
Revert "add adaptive buffer support for geo + device detection lookups"

elliottt:
Reverts fastly/Viceroy#383

This is a great change, but it introduces a mismatch in behavior with production. Let's reopen this as a PR that we can merge once we have support deployed in the fleet.


elliottt:
I totally jumped the gun here, and we do already have support in the fleet. Please ignore this one!

PR #386, opened 2024-06-13
Run the `request` tests as a component

elliottt:
Run request tests with components. This required changes to the long_header cases in `request.wasm`, as the strings passed need to be valid with the canonical abi.

### DIFF
--- a/lib/src/component/http_body.rs
+++ b/lib/src/component/http_body.rs
@@ -168,12 +168,17 @@ impl http_body::Host for Session {
             b'\0',
             usize::try_from(max_len).unwrap(),
             cursor,
-        );
-        if buf.is_empty() && next.is_none() {
-            return Ok(None);
-        }
+        )
+        .map_err(|needed| types::Error::BufferLen(u64::try_from(needed).unwrap_or(0)))?;
 
-        Ok(Some((buf, next)))
+        // At this point we know that the buffer being empty will also mean that there are no
+        // remaining entries to read.
+        if buf.is_empty() {
+            debug_assert!(next.is_none());
+            Ok(None)
+        } else {
+            Ok(Some((buf, next)))
+        }
     }
 
     async fn trailer_value_get(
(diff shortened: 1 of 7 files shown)


lib/src/component/http_req.rs
```diff
@@ -187,8 +181,17 @@ impl http_req::Host for Session {
...
         );
 
-        if buf.is_empty() && next.is_none() {
-            return Ok(None);
+        if buf.is_empty() {
+            if next.is_none() {
+                return Ok(None);
+            } else {
+                // It's an error if we couldn't write even a single value.
+                return Err(Error::BufferLengthError {
```
cee-dub, line 189: Does this allow us to return the needed size? I'm guessing not, since this is not the `BufferLen` variant.

lib/src/component/http_req.rs
```diff
@@ -236,8 +239,17 @@ impl http_req::Host for Session {
...
```
cee-dub, line 247: Same question about needed size

lib/src/component/http_req.rs
elliottt: Good point, I've fixed that by matching the behavior elsewhere: it returns the length needed to read at least one entry.

cee-dub: This seems like a nice ergonomic improvement!

lib/src/component/http_body.rs
```diff
@@ -168,12 +168,17 @@ impl http_body::Host for Session {
...
-            return Ok(None);
-        }
+        )
+        .map_err(|needed| types::Error::BufferLen(u64::try_from(needed).unwrap_or(0)))?;
 
-        Ok(Some((buf, next)))
+        // At this point we know that the buffer being empty will also mean that there are no
+        // remaining entries to read.
+        if buf.is_empty() {
+            debug_assert!(next.is_none());
```
acw, line 177: Do we have a test that runs over this case? Asserts always make me nervous.

acw (approved):
Seems reasonable!

lib/src/component/http_body.rs
elliottt: This was more just a guard against the future, for changes to `write_values`. I think it's impossible for `buf.is_empty()` to be true if `next.is_none()` is false, as that would be caught by early return in the loop in `write_values`. Additionally, writing a single empty entry into the result would cause at least the terminator byte `\0` to be written and the buffer to be non-empty.

PR #389, opened 2024-06-18
🎁 Populate nwritten_out when errors occur in config-store::get or dictionary::get

elliottt:
The wiggle representation of a function that returns `Result<u32, FastlyStatus>` and one that takes a `u32` outparam as its last argument and returns `Result<(), FastlyStatus>` are the same, which means that we can support communicating the length needed to resolve a `FastlyStatus::BufferLen` error without changing the abi of `fastly_config_store::get` or `fastly_dictionary::get`.

This PR makes this change to both `fastly_config_store::get` and `fastly_dictionary::get`, allowing SDKs to better adapt when buffer-length errors occur.

### DIFF
--- a/test-fixtures/src/bin/dictionary-lookup.rs
+++ b/test-fixtures/src/bin/dictionary-lookup.rs
@@ -10,9 +10,17 @@ fn main() {
             "animals".len(),
             &mut dict_handle as *mut _,
         );
-        assert_eq!(res, FastlyStatus::OK, "Failed to open dictionary");
+        assert_eq!(res, FastlyStatus::OK, "Failed to open config-store");
         dict_handle
     };
+
+    #[derive(Debug, PartialEq, Eq)]
+    enum LookupResult {
+        Success(String),
+        Missing,
+        BufTooSmall(usize),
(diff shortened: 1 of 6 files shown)


cee-dub (dismissed):
LGTM! I'm pretty new to reading Rust code, so if you want another reviewer, please pull someone in.

test-fixtures/src/bin/dictionary-lookup.rs
```diff
@@ -10,10 +10,18 @@ fn main() {
             "animals".len(),
             &mut dict_handle as *mut _,
         );
-        assert_eq!(res, FastlyStatus::OK, "Failed to open dictionary");
+        assert_eq!(res, FastlyStatus::OK, "Failed to open config-store");
```
lpereira, line 13: Maybe remove this file?

elliottt: I'll change it back to `dictionary`, thanks!

acfoltzer (approved):
Looks good, and matches the pending witx changes of the platform.

PR #390, opened 2024-06-18
🔧 Explicitly test the dictionary host calls in the dictionary fixture

elliottt:
The dictionary fixture is using the `ConfigStore` api from the fastly crate, which uses the config-store host calls under the hood. To ensure that we're testing the dictionary host calls while they exist, this PR refactors the dictionary-lookup fixture to use the dictionary host calls directly.


acw (approved):
LGTM!

PR #391, opened 2024-06-18
⁉️ Switch to using the on-demand allocator, instead of the pooling allocator

### LINKED ISSUE #255: Use the wasmtime on-demand allocator

acw:
Fixes #255.

In addition to just fixing #255, this also allows the test suite to be run in parallel, again. At least on my machines, running `make test` directly would cause memory failures. I'm guessing this was due to the pooling allocator's use of virtual memory. Switching to the on-demand allocator means that things run fine.

This extends the built-in `Limiter` object to use a `StoreLimits` under the hood, as suggested in the issue. It then builds different versions of this limiter based on whether or not Viceroy is operating in a component or non-component context, because component-based systems require more instances and more tables than traditional ones.

### DIFF
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -28,7 +28,10 @@ impl wasmtime::ResourceLimiter for Limiter {
         // gradually resize, this will track the total allocations throughout the lifetime of the
         // instance.
         self.memory_allocated += desired - current;
-        Ok(true)
+        // limit the amount of memory that an instance can use to (roughly) 128MB, erring on
+        // the side of letting things run that might get killed on Compute, because we are not
+        // tracking some runtime factors in this count.
+        Ok(self.memory_allocated < (128 * 1024 * 1024))
     }
 
     fn table_growing(
--- a/lib/src/linking.rs
+++ b/lib/src/linking.rs
@@ -6,15 +6,39 @@ use {
         wiggle_abi, Error,
     },
     std::collections::HashSet,
-    wasmtime::{GuestProfiler, Linker, Store, UpdateDeadline},
+    wasmtime::{GuestProfiler, Linker, Store, StoreLimits, StoreLimitsBuilder, UpdateDeadline},
     wasmtime_wasi::{preview1::WasiP1Ctx, WasiCtxBuilder},
     wasmtime_wasi_nn::WasiNnCtx,
 };
 
-#[derive(Default)]
 pub struct Limiter {
     /// Total memory allocated so far.
     pub memory_allocated: usize,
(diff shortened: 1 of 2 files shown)


elliottt:
Interestingly, we've always used the pooling allocator:

https://github.com/fastly/Viceroy/blob/f44257f62eb85cb39f768efa4f164472548ef911/lib/src/execute.rs#L74-L78

Should we integrate the `StoreLimitsBuilder` change that @acfoltzer suggested in this PR as well?

lib/src/linking.rs
```diff
@@ -6,38 +6,83 @@ use {
...
     ) -> anyhow::Result<bool> {
         // Track the diff in memory allocated over time. As each instance will start with 0 and
         // gradually resize, this will track the total allocations throughout the lifetime of the
         // instance.
         self.memory_allocated += desired - current;
-        Ok(true)
+        // limit the amount of memory that an instance can use to (roughly) 128MB, erring on
+        // the side of letting things run that might get killed on Compute, because we are not
+        // tracking some runtime factors in this count.
+        self.internal.memory_growing(current, desired, maximum)
```
elliottt, line 56: What do you think about incrementing `memory_allocated` only after a successful call to `self.internal.memory_growing(..)`? That way we would only report the memory that was successfully allocated.

elliottt (commented):
The `Limiter` changes look good to me, but I think we need a `StoreLimits` on `ExecuteCtx` as well, to ensure that we're enforcing limits on core wasm execution.

elliottt:
Ignore my comments about neeing a `Limiter` on `ExecuteCtx`, I missed that it's already present on `WasmCtx` :)

lib/src/linking.rs
```diff
@@ -90,7 +140,7 @@ impl ComponentCtx {
             wasi: builder.build(),
             session,
             guest_profiler: guest_profiler.map(Box::new),
-            limiter: Limiter::default(),
+            limiter: Limiter::new(100, 100),
```
elliottt, line 143: We'll need to change this, but I'm not too worried about it now as we're not officially supporting components yet.

elliottt (approved):
:ship:

PR #392, opened 2024-06-24
🎁 Update to wasmtime 22.0.0

elliottt:
Update to wasmtime 22.0.0.

This included some pretty substantial changes to the wiggle crate, which removes
runtime borrow checking of guest memory. This is a huge ergonomic improvement,
as the borrows are all tracked as rrust borrows instead.

One piece of functionality that's missing after the wiggle changes is the
ability to interact with a slice of guest memory that's not `&[u8]`. To work
around this, I'm copying memory over to the host, but in future releases of
wiggle we should have the ability to interact with slices that aren't just to
`u8`.

### DIFF
--- a/lib/src/async_io.rs
+++ b/lib/src/async_io.rs
@@ -26,7 +26,11 @@ impl FastlyAsyncIo for Session {
         let select_fut = self
             .select_impl(
                 memory
-                    // TODO: wiggle only supports slices of u8 in 22.0.0
+                    // TODO: `GuestMemory::as_slice` only supports guest pointers to u8 slices in
+                    // wiggle 22.0.0, but `GuestMemory::to_vec` supports guest pointers to slices
+                    // of arbitrary types. As `GuestMemory::to_vec` will copy the contents of the
+                    // slice out of guest memory, we should switch this to `GuestMemory::as_slice`
+                    // once it is polymorphic in the element type of the slice.
                     .to_vec(handles)?
                     .into_iter()
                     .map(|i| AsyncItemHandle::from(i).into()),
--- a/lib/src/async_io.rs
+++ b/lib/src/async_io.rs
@@ -7,28 +7,28 @@ use {
     futures::{FutureExt, TryFutureExt},
     std::time::Duration,
     tokio::time::timeout,
-    wiggle::GuestPtr,
+    wiggle::{GuestMemory, GuestPtr},
 };
 
 #[wiggle::async_trait]
 impl FastlyAsyncIo for Session {
-    async fn select<'a>(
+    async fn select(
         &mut self,
-        handles: &GuestPtr<'a, [AsyncItemHandle]>,
+        memory: &mut GuestMemory<'_>,
+        handles: GuestPtr<[AsyncItemHandle]>,
         timeout_ms: u32,
     ) -> Result<u32, Error> {
-        let handles = GuestPtr::<'a, [u32]>::new(handles.mem(), handles.offset())
-            .as_slice()?
-            .ok_or(Error::SharedMemory)?;
+        let handles = handles.cast::<[u32]>();
         if handles.len() == 0 && timeout_ms == 0 {
             return Err(Error::InvalidArgument);
         }
 
         let select_fut = self
             .select_impl(
-                handles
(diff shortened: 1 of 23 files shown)


lib/src/async_io.rs
```diff
@@ -7,28 +7,28 @@ use {
...
             return Err(Error::InvalidArgument);
         }
 
         let select_fut = self
             .select_impl(
-                handles
-                    .iter()
-                    .copied()
+                memory
+                    // TODO: wiggle only supports slices of u8 in 22.0.0
```
acw, line 29: This comment is confusing me a bit. Based on the comment, I would guess that `memory.to_vec(handles)` returns a `Vec<u8>`, which would be a problem because that would mean we'd be creating `AsyncItemHandle`s from bytes instead of 32-bit values. But if I look at the `wiggle` docs, it looks like `memory_to_vec(handles)` is going to keep the type `T` (in this case, `u32`), so we'll be doing the right thing here.

So ... what's up?

lib/src/async_io.rs
```diff
@@ -7,28 +7,28 @@ use {
...
+                    .to_vec(handles)?
+                    .into_iter()
                     .map(|i| AsyncItemHandle::from(i).into()),
```
acfoltzer, line 32: Is there a limitation implied here, like we can only support 256 async item handles rather than the full u32 range of handles previously supported?

acfoltzer: OH, is the limitation that we can only have _slices_ of u8s, but we can have full `Vec`s of other scalar types?

acfoltzer (dismissed):
There's quite a few lines in this diff, but they're pleasantly straightforward and mechanical. I had feared this would be a more disruptive transition than it actually turned out to be, and I'm really happy to have more of the ownership properties now checked statically.

The only thing I might be tempted to change here is the comment in the code about only supporting slices of `u8`. It's a bit ambiguous to the reader, and it might cause some surprise until they figure out that we're using a `Vec` instead of a slice rather than a `u8` instead of some other type.

lib/src/wiggle_abi/req_impl.rs
```diff
@@ -772,39 +821,41 @@ impl FastlyHttpReq for Session {
...
         // perform the select operation
         let done_index = self
             .select_impl(
-                pending_req_handles
-                    .as_slice()?
-                    .ok_or(Error::SharedMemory)?
-                    .iter()
-                    .map(|handle| PendingRequestHandle::from(*handle).into()),
+                memory
+                    // TODO: wiggle only supports slices of u8 in 22.0.0
```
acw, line 849: See previous TODO comment.

acw (changes requested):
I think there are two comments that could use at least a bit more explanation, but otherwise this looks good. I tried to be pretty careful in looking at the diffs, but I have to admit that my eyes starting glazing over at some points. I can confirm that the tests still pass locally, though.

lib/src/async_io.rs
elliottt: Yep, that's exactly it. I'm going to fix wiggle so that we can have slices of other things, and this comment was mostly a reminder to me that this should be updated to not copy in the future.

lib/src/async_io.rs
elliottt: Yep, this needs to be reworded. The issue is that the `as_slice` function on `GuestMemory` will only handle `GuestPtr<[u8]>`, and not `GuestPtr<[T]>`. This was just an oversight, and in the future we can avoid copying with `to_vec` here.

elliottt: (Just to be really clear: this does the right thing when producing a vector of `u32`, but it copies in the process)

lib/src/wiggle_abi/req_impl.rs
elliottt: Updated!

elliottt:
I've updated the comment to hopefully clarify the situation with the use of `GuestMemory::to_vec`.

acw (approved):
Yay!

acfoltzer (approved):
I am pleased to see the Adams are in alignment on this PR

PR #393, opened 2024-06-26
🔨 Populate the guest cli args

### LINKED ISSUE #376: Golang programs that import flag or httptest will panic due to uninitialized os.Args
When I try to run `fastly compute serve` with webassembly code that I build with `GOARCH=wasm GOOS=wasip1 go build  ./...` if I have code that even indirectly includes `flag` then the `flag` package imports `httptest  ( _ "net/http/httptest")` which will panic when it gets to:
https://github.com/golang/go/blob/3ed007d754b685cd8f6011a8e96a3c9303c785db/src/flag/flag.go#L1199

An example piece of code 
```go
package main

import (
	"fmt"
	"os"

	_ "bytes"
	_ "context"
	_ "encoding/json"
	_ "net"
	_ "net/http"

	// This causes issues!
	_ "net/http/httptest"

	_ "net/url"
	_ "strconv"
	_ "strings"
)

func main() {
	fmt.Println("FASTLY_SERVICE_VERSION:", os.Getenv("FASTLY_SERVICE_VERSION"))
}
```
and run `fastly compute serve`
with this `fastly.toml` file:

```toml
# This file describes a Fastly Compute package. To learn more visit:
# https://www.fastly.com/documentation/reference/compute/fastly-toml

authors = ["XXXXacademy.org"]
cloned_from = "https://github.com/fastly/compute-starter-kit-go-default"
description = "fastly compute go service"
language = "go"
manifest_version = 3
name = "fastly-khanacademy-compute"
service_id = "XXXX"

[scripts]
  build = "go build -o bin/main.wasm ./cmd/serve"
  env_vars = ["GOARCH=wasm", "GOOS=wasip1"]
  post_init = "go get github.com/fastly/compute-sdk-go@latest"

# Despite most of the local_server configuration being in the fastly.{env}.toml
# files, this actually needs to be defined in our basic fastly.toml file.
# It MUST be defined as the full semantic version, but without the 'v'.
# The 'fastly' cli must be at version 10.2.0 or later to use this.
[local_server]
  viceroy_version = "0.9.6"
```  
I can reproduce this problem with tinygo, but wasmtime doesn't seem to have the problem directly, so I assume it is a viceroy problem.

I mentioned this in the gophers slack and @dgryski said he passed it on to a colleague, but I didn't want others to duplicate bug reports, so I thought I'd record it here for others to track.

elliottt:
Populate the guest cli args with the single value `"compute-app"`. This should simplify adoption for languages that assume the args list is non-empty.

Fixes #376

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -454,8 +454,10 @@ impl ExecuteCtx {
                 let req = session.downstream_request();
                 let body = session.downstream_request_body();
 
-                let mut store = ComponentCtx::create_store(&self, session, None, |_| {})
-                    .map_err(ExecutionError::Context)?;
+                let mut store = ComponentCtx::create_store(&self, session, None, |ctx| {
+                    ctx.arg("compute-app");
+                })
+                .map_err(ExecutionError::Context)?;
 
                 let (compute, _instance) =
                     compute::Compute::instantiate_pre(&mut store, instance_pre)
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -511,8 +511,10 @@ impl ExecuteCtx {
                 // due to wasmtime limitations, in particular the fact that `Instance` is not `Send`.
                 // However, the fact that the module itself is created within `ExecuteCtx::new`
                 // means that the heavy lifting happens only once.
-                let mut store = create_store(&self, session, profiler, |_| {})
-                    .map_err(ExecutionError::Context)?;
+                let mut store = create_store(&self, session, profiler, |ctx| {
+                    ctx.arg("compute-app");
+                })
+                .map_err(ExecutionError::Context)?;
 
                 let instance = instance_pre
                     .instantiate_async(&mut store)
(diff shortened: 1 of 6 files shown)


lib/src/execute.rs
```diff
@@ -511,8 +511,10 @@ impl ExecuteCtx {
                 // due to wasmtime limitations, in particular the fact that `Instance` is not `Send`.
                 // However, the fact that the module itself is created within `ExecuteCtx::new`
                 // means that the heavy lifting happens only once.
-                let mut store = create_store(&self, session, profiler, |_| {})
-                    .map_err(ExecutionError::Context)?;
+                let mut store = create_store(&self, session, profiler, |ctx| {
+                    ctx.arg("compute-app");
```
jameysharp, line 515: Do you also want to add this for the `ComponentCtx::create_store` case as well?

jameysharp (approved):
Nice, I like that this makes `viceroy serve` closer to `viceroy run`, which uses the stem of the `.wasm` filename as the program name.

lib/src/execute.rs
elliottt: Good catch!

elliottt: I've added this for components and enabled testing for it, but the adapter will need some changes to get the test passing. I'll put that on my TODO list for component parity, but don't think the adapter changes should block this PR.

jameysharp (approved):

PR #394, opened 2024-06-28
Support cli args in the adapter

elliottt:
Integrate the adapter allocator changes from https://github.com/bytecodealliance/wasmtime/pull/8594, which simplifies supporting cli args.

Additionally, enable the cli args test for components, as it's now passing.


cli/tests/integration/args.rs
```diff
@@ -29,7 +29,6 @@ async fn empty_ok_response_by_default_after_args() -> TestResult {
 // TODO: The adapter needs to plumb through support for argument handling. This was removed
```
kpfleming, line 29: This comment is probably out of date now.

jameysharp (approved):
I didn't review the memory management commit carefully, on the assumption that you copied that code directly from Wasmtime where it's already been reviewed. But the command-line args changes look sensible to me!

PR #395, opened 2024-06-28
Rework component testing support to make test updates easier

elliottt:
Add a macro for generating the core wasm and component versions of a test, and
modify the `.adapt_component` builder function on `Test` to accept a bool. This
simplifies modifying existing tests to support components by doing the work of
abstracting out the body in a macro.


jameysharp (approved):
Yup, this strikes me as a net positive for test clarity, in spite of my general preference to avoid macros. :+1:

PR #396, opened 2024-06-28
Run tests as components

elliottt:
Update tests to run as both core-wasm modules and components. This also includes a number of fixes to both the component adapter and component host side implementation.


cli/tests/integration/common.rs
```diff
@@ -48,6 +48,29 @@ macro_rules! viceroy_test {
...
+/// little bit harder to debug, but for tests like logging that manipulate a shared resource, this
+/// is the simplest solution.
+#[macro_export]
+macro_rules! viceroy_test_sequential {
+    ($name:ident, |$is_component:ident| $body:block) => {
+        mod $name {
+            use super::*;
+
+            async fn test_impl($is_component: bool) -> TestResult {
+                $body
```
acw, line 62: OK, I had expected the comment above to point me in the direction of some shared state in $body that made us want to run things in a single function like below. But given this, the concern is shared global state that is manipulated by $body?

If that's true, can I suggest taking a look at the `serial_test` crate? If I'm remembering it correctly, and if $name is unique (or at least mostly unique), you could then duplicate the test using:

```
#[tokio::test(flavor = "multi_thread")]
#[serial($name)]
async fn noncomponent() -> TestResult {
    test_impl(false).await?
}

#[tokio::test(flavor = "multi_thread")]
#[serial($name)]
async fn component() -> TestResult {
    test_impl(true).await?
}
```

which would allow us to know which of the two tests failed, which would be really nice.

acw (changes requested):
I'd like to see if the `serial` thing can work out, to help us understand what's breaking. But if it doesn't, then ping me and I'll approve without it.

acw (approved):
Yay testing!

PR #397, opened 2024-06-28
Allow capturing logging endpoint messages

jameysharp:
Some integration tests need to capture log output in order to verify that logging endpoints work correctly from guests, but it's also potentially useful for people using Viceroy as a library.

The previous way that the integration tests did this was not reliable when multiple tests ran in parallel, because there was only one global hook for capturing logs. Exposing this as a configuration option on execution contexts instead allows each test to set independent capture buffers.

I guess this should be considered a breaking API change since `viceroy_lib::logging::LOG_WRITER` was exported publicly from the crate.

### DIFF
--- a/lib/src/logging.rs
+++ b/lib/src/logging.rs
@@ -1,26 +1,23 @@
-use {
-    lazy_static::lazy_static,
-    std::{
-        io::{self, Write},
-        sync::Mutex,
-    },
+use std::{
+    io::{self, Write},
+    sync::{Arc, Mutex},
 };
 
-/// A logging endpoint, which for Viceroy is just a name.
-pub struct LogEndpoint(Vec<u8>);
-
-lazy_static! {
-    /// The underlying writer to use for all log messages. It defaults to `stdout`,
-    /// but can be redirected for tests. We make this a static, rather than e.g.
-    /// a field in `ExecuteCtx`, because the `Write` implementation for `LogEndpoint`
-    /// doesn't have direct access to context data.
-    pub static ref LOG_WRITER: Mutex<Box<dyn Write + Send>> = Mutex::new(Box::new(io::stdout()));
+/// A named logging endpoint.
+#[derive(Clone)]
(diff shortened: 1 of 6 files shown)


lib/src/logging.rs
```diff
@@ -1,26 +1,22 @@
...
-lazy_static! {
-    /// The underlying writer to use for all log messages. It defaults to `stdout`,
-    /// but can be redirected for tests. We make this a static, rather than e.g.
-    /// a field in `ExecuteCtx`, because the `Write` implementation for `LogEndpoint`
-    /// doesn't have direct access to context data.
-    pub static ref LOG_WRITER: Mutex<Box<dyn Write + Send>> = Mutex::new(Box::new(io::stdout()));
+#[derive(Clone)]
+pub struct LogEndpoint {
+    name: Vec<u8>,
+    writer: Arc<Mutex<dyn Write + Send>>,
```
elliottt, line 10: Since this escapes out now, would it be worth adding a synonym for `Arc<Mutex<dyn Write + Send>>` for readability elsewhere?

elliottt (approved):
That global mutex has bothered me for a long time, thank you for fixing this!

lib/src/logging.rs
jameysharp: Good question. While trying to decide how to answer it, I've stared at the places this is visible, and I think my answer is that it shouldn't actually be visible in most of those places. In particular, I don't think either the `LogEndpoint` or `Session` types should be exported. The only place library users should encounter this type is in the `ExecuteCtx::with_capture_logs` method, and I don't think a type alias is necessary for that.

It's a little annoying naming this type multiple places within the library, but I think it's a little better to keep the details visible so that, for example, it's clear that using `clone` is cheap.

So I guess I'm not inclined to introduce a type alias for this.

elliottt: Reasonable!

elliottt (approved):

PR #398, opened 2024-06-28
Remove some unused memory management code in the adapter

elliottt:
After #394 there were some unused functions in the adapter that didn't fail in
CI as we don't currently build the adapter in CI. This PR removes the unused
code, and also builds the adapter in CI.


jameysharp (dismissed):
Because rustc has introduced new warnings from time to time, I'm always hesitant about introducing blanket `deny(warnings)` as it can mean that upgrading rustc breaks the build. I'm not sure whether to consider that a significant issue here; up to you!

Otherwise, this is clearly sensible. Ship it!

jameysharp (approved):

PR #399, opened 2024-07-01
Support `wat` files when adapting core wasm

elliottt:
Support `.wat` files when adapting with either the `--adapt` flag, or the `viceroy adapt` command by first turning them into binary modules. Also, enable the tests that use wat files directly for components.


acw (approved):
LGTM!

acw (approved):
:recycle:

PR #400, opened 2024-07-02
Add support for environment variables in the adapter

elliottt:
This copies over the implementation of `environ_get` and `environ_sizes_get` from the upstream adapter in wasmtime, and enables the environment variable test for components.


jameysharp (approved):
Seems reasonable to me and it's nice to get more of the component tests running!

PR #401, opened 2024-07-03
Add `downstream_server_ip_addr` hostcall

ulyssa:
A customer noticed when trying to use the Rust SDK's `Request::get_server_ip` method that the  `downstream_server_ip_addr` hostcall is not yet implemented in Viceroy. This PR adds support for it. I've also updated the SDK version used in the tests to be the latest release.


elliottt (approved):
Looks great!

PR #403, opened 2024-07-03
Add `downstream_compliance_region` hostcall

ulyssa:
This PR implements the `downstream_compliance_region` hostcall, which will eventually be exposed in the SDK when the Compliance Regions GA happens. I've made it so that it currently always returns `"none"`, which is what you will see in Compute when you aren't using Compliance Regions with your service.

I've also updated the signature of the component interface to be more similar to `downstream-tls-ja3-md5`, since it just returns a `Vec<u8>` and doesn't need to know the `max-len`. (Let me know if I'm misunderstanding how the component interfaces should work.)

### DIFF
--- a/lib/src/component/http_req.rs
+++ b/lib/src/component/http_req.rs
@@ -808,8 +808,17 @@ impl http_req::Host for Session {
         Err(Error::NotAvailable("Client TLS JA4 hash").into())
     }
 
-    async fn downstream_compliance_region(&mut self) -> Result<Vec<u8>, types::Error> {
-        Ok(Vec::from(b"none"))
+    async fn downstream_compliance_region(
+        &mut self,
+        region_max_len: u64,
+    ) -> Result<Vec<u8>, types::Error> {
+        let region = Session::downstream_compliance_region(self);
+        let region_len = region.len();
+
+        match u64::try_from(region_len) {
+            Ok(region_len) if region_len <= region_max_len => Ok(region.into()),
+            too_large => Err(types::Error::BufferLen(too_large.unwrap_or(0))),
+        }
     }
 
     async fn original_header_names_get(
--- a/lib/src/component/http_req.rs
+++ b/lib/src/component/http_req.rs
@@ -808,11 +808,8 @@ impl http_req::Host for Session {
         Err(Error::NotAvailable("Client TLS JA4 hash").into())
     }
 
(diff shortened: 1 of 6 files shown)


lib/wit/deps/fastly/compute.wit
```diff
@@ -323,7 +323,7 @@ interface http-req {
 
   downstream-tls-ja4: func(max-len: u64) -> result<list<u8>, error>;
 
-  downstream-compliance-region: func(max-len: u64) -> result<list<u8>, error>;
+  downstream-compliance-region: func() -> result<list<u8>, error>;
```
elliottt, line 326: Could you add `max-len` back in? It's to help with the adapter, as it's ensuring that the returned vector that will be placed in the guest-provided buffer will be large enough.

lib/src/component/http_req.rs
```diff
@@ -808,11 +808,8 @@ impl http_req::Host for Session {
         Err(Error::NotAvailable("Client TLS JA4 hash").into())
     }
 
-    async fn downstream_compliance_region(
-        &mut self,
-        _max_len: u64,
-    ) -> Result<Vec<u8>, types::Error> {
-        Err(Error::NotAvailable("Client TLS JA4 hash").into())
+    async fn downstream_compliance_region(&mut self) -> Result<Vec<u8>, types::Error> {
+        Ok(Vec::from(b"none"))
```
elliottt, line 812: Could you also add a check here that `max_len >= 4`?

elliottt (commented):
Looks good, though the `max_len` param should be kept if `downstream_compliance_region` can return an arbitrary-length vector.

kpfleming:
Can we find a place to keep track of hardcoded values like this, so that as part of the discussion about making development and testing easier (the 'redesign fastly.toml' topic) we can provide ways to make these configurable in the future? "none" is certainly a valid and expected value, but customers should be able to test their code against the other values they could see in production.

cc @kailan

lib/wit/deps/fastly/compute.wit
ulyssa: Done in 656a5a9.

lib/src/component/http_req.rs
ulyssa: Done in 656a5a9.

elliottt (approved):
Looks good to me!

PR #404, opened 2024-07-08
🎁 Prepare release v0.10.0

ulyssa:
This updates the lockfiles with `make generate-lockfile` as mentioned in [RELEASING.md](https://github.com/fastly/Viceroy/blob/main/doc/RELEASING.md), and updates the `CHANGELOG.md` to include the commits added since #375.

Rendered view of the changelog is [here](https://github.com/fastly/Viceroy/blob/release-0.9.8/CHANGELOG.md#098-2024-07-08).


acfoltzer (changes requested):
It looks like this needs to be a major version bump to 0.10.0, as the `viceroy-lib` public interface includes types from `wasmtime` which got a major bump. There are also a couple of public definitions within the crate itself whose types have changed.

Given the large surface area of exports from the `viceroy-lib` crate, we might want to consider a policy similar to Wasmtime of just making a major bump on every release in order to avoid accidental semver breaks. Or making many more things private to limit the surface area we have to audit when deciding on the next release's version.

ulyssa:
That makes sense! I've bumped the `Cargo.toml` files to use `0.10.0`, and updated the `CHANGELOG.md` to reflect the new version number. (I can't change the source branch of the PR, but I can close and open a new one if the branch name is important; it looks like previous PRs automatically deleted the branch, so only the tag will matter, I think.)

acfoltzer (approved):
+1 to release per step 6 of the release process 🚀

acfoltzer:
> (I can't change the source branch of the PR, but I can close and open a new one if the branch name is important; it looks like previous PRs automatically deleted the branch, so only the tag will matter, I think.)

Yeah, the tag name is what counts

ulyssa:
It looks like the `release.yml` may need the same `test.yml` changes made in #398 in order to build the component adapter.

acfoltzer:
Hmm, glad you found a workaround, but flagging this for @elliottt when he's back. There's definitely some awkwardness about having the adapter where it currently sits.

ulyssa:
02a15e6 adds `## Unreleased` back to the top of `CHANGELOG.md` and updates the version to v0.10.1 for now. If we want to make each release a breaking change going forward, we can update it to `0.11.0` on the next release.

acfoltzer (approved):
LGTM; thank you for navigating this unusual release!

PR #412, opened 2024-07-29
⏱️ At a hostcall to get the amount of vcpu time that has passed in the guest, in milliseconds

acw:
This is a proposed extension, to help customers diagnose where their code is spending time.

One explicit thing to mention is that while this hostcall limits itself to vcpu time (and thus does not count any time spent blocking on IO, for example), it is based on a time passing rather than cycles, WASM instruction count, or any other timer that is independent of the underlying processor(s) running the code. Thus, the values found should only really be compared to other values captured in the exact same run.


crates/adapter/src/fastly/core.rs
```diff
@@ -268,6 +268,24 @@ pub mod fastly_abi {
     }
 }
 
+pub mod fastly_vcpu {
```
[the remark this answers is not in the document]
aturon: (I can imagine a lot of additional introspection we might want to add over time, on other aspects of the guest behavior.)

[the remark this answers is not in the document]
acw: I went with `fastly_compute_runtime`, which I like as a general future module for getting or setting or otherwise looking at the behavior of a session itself.

lib/src/session.rs
```diff
@@ -47,6 +48,8 @@ pub struct Session {
     downstream_client_addr: SocketAddr,
     /// The IP address and port that received this session.
     downstream_server_addr: SocketAddr,
+    /// The amount of time we've spent on this session in ms.
+    pub active_cpu_time_us: Arc<AtomicU64>,
```
cee-dub, line 52:
```suggestion
    /// The amount of time we've spent on this session in microseconds.
    pub active_cpu_time_us: Arc<AtomicU64>,
```

Was reading this PR out of curiosity and to practice reading Rust and noticed these units mismatch.

acw: Yes, they do! Thanks for the catch!

lib/data/viceroy-component-adapter.wasm
ulyssa, line 1: Did you rerun `make adapter` after merging main so that this includes both the vcpu hostcall and the inspect hostcall?

ulyssa (approved):
This looks good to me! :sparkles:  Just the question about the adapter wasm.

lib/data/viceroy-component-adapter.wasm
acw: Yup! Rerunning it now shows me no changes with this PR.

PR #414, opened 2024-08-13
Update compute.wit and the adapter for some api fixes

elliottt:
While implementing compute.wit elsewhere, I made a few changes that better reflect the translation to the underlying hostcalls. This PR backports those changes to compute.wit, and folds them in to the adapter and component implementation.


acw (approved):
Oh, I like some of these much better.

PR #415, opened 2024-08-13
Rename the kv-store interface to object-store in compute.wit

elliottt:
Rename the `kv-store` interface in compute.wit to `object-store`, and add `todo!()` implementations for the new `kv-store` interface. The host implementations of those functions will be added in a follow-up PR along with the wiggle implementation for the new host calls, however the adapter has been updated to handle the new hostcalls already.

I tried to structure the commits so that it shows the renaming and updates to `object_store.rs` as an actual diff, but github is not cooperating. Looking at the individual commits should show that the changes to `object_store.rs` includes a renaming and small changes to the original implementation, while `kv_store.rs` is a completely fresh implementation of a new interface.

cc @computermouth


lib/wit/deps/fastly/compute.wit
```diff
@@ -711,6 +711,110 @@ interface kv-store {
   ) -> result<_, error>;
 }
 
+/*
+ * Fastly KV Store
+ */
+interface kv-store {
```
acw, line 717: Is there any sort of deprecation tag (or even a comment) we can put here?

acw (approved):
Looked right, but I'd love it if there were some way we could mark the kv store wit as deprecated.

lib/wit/deps/fastly/compute.wit
elliottt: There is, but we could instead just remove the `object-store` interface and polyfill it in the adapter with the new `kv-store` import. Either way, let's hold off until the `kv-store` interface has been filled out :)

PR #417, opened 2024-08-16
Implement stub for `inspect` hostcall

ulyssa:
This adds a small stub for the `inspect` hostcall that makes sure its input looks okay and then always returns an allow response. I've put the `"allow"` value into the `Session` so that in the future it should be possible to make it configurable, so that folks can change it to `block` when testing denied requests.


crates/adapter/src/fastly/core.rs
```diff
@@ -1689,6 +1689,27 @@ pub mod fastly_http_req {
...
+        buf: *mut u8,
+        buf_len: usize,
+        nwritten_out: *mut usize,
+    ) -> FastlyStatus {
+        alloc_result!(buf, buf_len, nwritten_out, {
+            fastly::api::http_req::inspect(
+                ds_req,
+                ds_body,
+                info_mask,
+                info,
```
elliottt, line 1708: I think that `info` will need to be rebuilt as a `fastly::api::http_req::InspectConfig`. There's an example of the same over here: https://github.com/fastly/Viceroy/blob/a945b345b422df1c02db115368d44bafe7882f7a/crates/adapter/src/fastly/core.rs#L1460-L1488

elliottt (commented):
This looks great! There's only one change I'd recommend for the adapter, which is to rebuild the flags and config values for the wit-generated types.

crates/adapter/src/fastly/core.rs
ulyssa: Thank you for pointing me to that! I've set up something similar for `InspectConfigOptions` and `InspectConfig` in af24988220b52134492a93d88e44727d3e9e6292.

elliottt (approved):
This looks great!

PR #418, opened 2024-08-20
📦 Prepare v0.11.0 release

ulyssa:
Updates lockfiles and `CHANGELOG.md` per the [release instructions](https://github.com/fastly/Viceroy/blob/main/doc/RELEASING.md).

(I've bumped this to `0.11.0` since it looks like #414 would count as a breaking change. lmk if we should do an `0.10.3` release.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
-## Unreleased
+## 0.11.0 (2024-08-20)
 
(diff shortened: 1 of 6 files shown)


elliottt (dismissed):
Looks great!

kailan (approved):
Looks good ✨

PR #423, opened 2024-08-30
Add keepalive options for dynamic backends

ulyssa:
This adds the upcoming dynamic backend options for configuring HTTP and TCP keepalive parameters. They aren't available in Compute yet, so I'm keeping this a draft for right now.


elliottt (approved):
Looks good!

PR #424, opened 2024-09-03
Add ReplaceHandle hostcall stubs for upcoming SDK release

ulyssa:
The upcoming Rust SDK release will include an API for using `replace_*` variants of a number of the existing cache API hostcalls that operate on a `ReplaceHandle`. Since Viceroy doesn't handle any of the caching APIs for now, this PR only adds stubs for these hostcalls and returns an error just like the existing ones do.


acw (approved):
LGTM!

PR #428, opened 2024-09-16
add support for reading secrets from a JSON file

### LINKED ISSUE #364: Allow JSON files as input for Secrets Stores and KVStores

joeshaw:
This uses the same format as for the KV store, introduced in #365.

Fixes #364.

### DIFF
--- a/lib/src/config/secret_store.rs
+++ b/lib/src/config/secret_store.rs
@@ -23,10 +23,10 @@ impl TryFrom<Table> for SecretStoreConfig {
                 });
             }
 
-            // Either the items here is from a top-level file with
+            // Either the items here are from a top-level file with
             // "file" and "format" keys or it's an inline array.
             // We try to parse either one of them to the same Vec<toml::Value>
-            // to allow them to run through the same validation path futher down
+            // to allow them to run through the same validation path further down.
             let file_path = items
                 .as_table()
                 .and_then(|table| table.get("file"))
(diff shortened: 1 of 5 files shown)


lib/src/config/secret_store.rs
```diff
@@ -23,12 +23,66 @@ impl TryFrom<Table> for SecretStoreConfig {
                 });
             }
 
-            let items = items.as_array().ok_or_else(|| {
-                FastlyConfigError::InvalidSecretStoreDefinition {
-                    name: store_name.to_string(),
-                    err: SecretStoreConfigError::NotAnArray,
+            // Either the items here is from a top-level file with
```
cee-dub, line 26:
```suggestion
            // Either the items here are from a top-level file with
```

lib/src/config/secret_store.rs
```diff
@@ -23,12 +23,66 @@ impl TryFrom<Table> for SecretStoreConfig {
...
+            // "file" and "format" keys or it's an inline array.
+            // We try to parse either one of them to the same Vec<toml::Value>
+            // to allow them to run through the same validation path futher down
```
cee-dub, line 29:
```suggestion
            // to allow them to run through the same validation path further down.
```

cee-dub (commented):
Just a couple spelling corrections from me. I didn’t see any obvious issues with the Rust, but I’m still learning it, so I don’t feel qualified to give a ✅.

ulyssa (approved):
This looks good to me! :sparkles:

PR #433, opened 2024-10-02
Stub new HTTP cache hostcalls (witx only)

aturon:
Adds the hostcalls for the forthcoming HTTP cache API, stubbed as "not available".

This will enable releasing the Rust SDK using these features. For the sake of time, the Component bindings are left for follow up.


aturon:
I've manually verified that code compiled against the new SDK with HTTP caching will run against this Viceroy.

acw (approved):
LGTM!

PR #434, opened 2024-10-04
Fix the alignment check in the adapter's allocator

elliottt:
Integrate https://github.com/bytecodealliance/wasmtime/pull/9356 into the adapter, to fix allocations of zero-bytes from a zero-byte buffer.


acw (approved):
I hate > vs >=

PR #436, opened 2024-10-11
Remove macOS 12 (soon to be unavailable), add macOS 15 (beta runner).

kpfleming:


kpfleming:
Once this is merged I'll adjust the branch protection settings to require macOS 15 and no longer require macOS 12.

ulyssa (approved):

PR #437, opened 2024-10-12
Enable custom profiling interval via CLI flag (#357)

omarjatoi:
**Issue:** #357 (guest profiler support originally added in https://github.com/fastly/Viceroy/pull/301)

This PR adds the ability to pass a profiling interval to Viceroy, e.g. `--profile=guest,/tmp/path,25` will set the profiling sample interval to 25 ms.

**Open question(s):**

1. ~~It looks like `wasmtime` will parse the time including units, e.g. `--profile=guest,/some/path,25ms`. This PR assumes milliseconds by default. I can either add a small `parse_duration` helper function or use the [`humantime` crate](https://docs.rs/humantime/latest/humantime/) ([used by `wasmtime`](https://github.com/bytecodealliance/wasmtime/blob/460a4c0ad5d3655b7dbaec193e726196de848898/crates/cli-flags/Cargo.toml#L22), so already an implicit project dependency) to stay consistent with `wasmtime`. Thoughts?~~ 
**EDIT:** I've added support for `humantime`, but if this is not preferable I'm happy to revert the latest change.
2. In Viceroy, the [default](https://github.com/fastly/Viceroy/blob/274230fcfebde926666d78ed27e0f34e4ff3f09e/lib/src/execute.rs#L48) sampling interval is [set](https://github.com/fastly/Viceroy/blob/274230fcfebde926666d78ed27e0f34e4ff3f09e/lib/src/execute.rs#L641) to be `0.05` ms (50 microseconds). The default interval in `wasmtime` [appears to be `10` ms](https://github.com/jameysharp/wasmtime/blob/main/docs/examples-profiling-guest.md). Should Viceroy's default be changed?
3. Current CLI tests do not test anything related to profiling, I'm not sure how valuable those tests would be but I can certainly try to write something to cover the functionality/changes in this PR.


omarjatoi:
Hi @katef, I moved this from draft to 'ready for review'. Please let me know if this looks alright!

omarjatoi:
Looks like this was completed in https://github.com/fastly/Viceroy/pull/456 🤷, closing this PR.

posborne:
> Looks like this was completed in https://github.com/fastly/Viceroy/pull/456, closing this PR.

Yes, thank you for submitting this; I happened to miss that it existed before having #456 implemented otherwise I would have definitely gone the route of getting this review and merged.  I stumbled upon the lack of this feature (which can be very useful) independently while doing some profiling.

omarjatoi:
No worries @posborne, thanks for the note!

PR #438, opened 2024-10-24
Add `fastly_acl` hostcalls

awilliams-fastly:
This adds support for two new hostcalls:
- `fastly_acl::open`
- `fastly_acl::lookup`

ACLs can be mocked via `fastly.toml` by adding one or more `acls` to `local_server`. Each acl entry is a reference to an external JSON file. Example:

```toml
[local_server]
acls.acl-v4 = "./acl-v4.json"
acls.acl-v6 = {file = "./acl-v6.json"}
```

The JSON file contains a list of one or more ACL entries, in the same format as can be used to modify ACLs via api.fastly.com. Example:

```json
{
  "entries": [
    { "prefix": "1.2.3.0/24", "action": "BLOCK" },
    { "prefix": "192.168.0.0/16", "action": "BLOCK" },
    { "prefix": "23.23.23.23/32", "action": "ALLOW" },
    { "prefix": "1.2.3.4/32", "action": "ALLOW" },
    { "prefix": "127.0.0.0/8", "action": "ALLOW" }
  ]
}
```

The [test-fixtures/src/bin/acl.rs](https://github.com/fastly/Viceroy/blob/6aae133d0a75cf423064b07c40ccce8830ee7e98/test-fixtures/src/bin/acl.rs) file shows example guest usage.

### Integration Tests
The `fastly` Rust crate containing support for ACLs has not yet been released. Because of that, the `acl.rs` integration test uses a configuration feature to conditionally compile, with a default of not using the `fastly::acl` feature:
https://github.com/fastly/Viceroy/blob/6aae133d0a75cf423064b07c40ccce8830ee7e98/test-fixtures/src/bin/acl.rs#L5-L7

To test against a local, unreleased crate:
- Modify `test-fixtures/Cargo.toml` to point to the appropriate local crates
- Compile with `acl_hostcalls` feature: `(cd test-fixtures && cargo build -F acl_hostcalls --target=wasm32-wasi)`

Once a new version of the crate has been published, the conditional can be removed. I have tested against the internal/unreleased crate and verified that it passes.


test-fixtures/src/bin/acl.rs
```diff
@@ -0,0 +1,62 @@
+//! A guest program to test that acls works properly.
+use fastly::Error;
+
+fn main() -> Result<(), Error> {
+    // Temporary until fastly SDK is released which
+    // includes the fastly::acl module.
+    #[cfg(feature = "acl_hostcalls")]
```
ulyssa, line 7: If I'm following right, is the idea here to:

- Merge this PR
- Publish Viceroy
- Publish Rust SDK with ACLs support
- Do follow-up PR to remove the `acl_hostcalls` feature guard

lib/src/config/acl.rs
```diff
@@ -0,0 +1,67 @@
+use crate::acl;
+
+#[derive(Clone, Debug, Default)]
+pub struct AclConfig(pub(crate) acl::Acls);
+
+mod deserialization {
```
ulyssa, line 6: It seems fine, but I'm curious: is there a reason to put this into its own `deserialization` module? And would it be worth putting it into its own `lib/src/config/acl/deserialization.rs`?

ulyssa (dismissed):
I think this looks good! I just have two minor questions.

test-fixtures/data/my-acl-1.json
```diff
@@ -0,0 +1,8 @@
+{
```
mhp, line 1: I really like the fact that the file formats are close to the format used to send updates to the API.  How do you feel about adding the [`"op"` field](https://github.com/fastly/api-planning/blob/main/proposals/new-apis/acls_in_compute.md#input-json-parameters-1) too, so that the formats are the same?

lib/src/acl.rs
```diff
@@ -0,0 +1,386 @@
...
+
+#[test]
+fn acl_json_parse() {
+    let input = r#"
+    { "entries": [
+        { "op": "create", "prefix": "1.2.3.0/24", "action": "BLOCK" },
+        { "op": "update", "prefix": "192.168.0.0/16", "action": "BLOCK" },
+        { "op": "create", "prefix": "23.23.23.23/32", "action": "ALLOW" },
+        { "op": "update", "prefix": "1.2.3.4/32", "action": "ALLOW" },
+        { "op": "update", "prefix": "1.2.3.4/8", "action": "ALLOW" }
```
mhp, line 282: It looks like this testcase includes the `"op"` field, which is nice.  Is it just ignored then?  That's probably fine then!

lib/src/acl.rs
```diff
@@ -0,0 +1,386 @@
...
+pub struct Acl {
+    pub(crate) entries: Vec<Entry>,
+}
+
+impl Acl {
+    /// Lookup performs a naive lookup of the given IP address
+    /// over the acls entries.
+    ///
+    /// If the IP matches multiple ACL entries, then:
+    /// - The most specific match is returned (longest mask),
```
mhp, line 40: This looks like it gives the same results as we'd expect to see in production. Obviously the performance characteristics will be different, but I don't believe that it is a goal of Viceroy to duplicate that.

mhp (approved):
Thanks for doing this work.  I've left a couple of comments regarding the json format, but I think I just convinced myself it was probably fine.  I've also left a note to the effect that the goal here is to match production behaviour on a functional basis, not on a performance basis.

test-fixtures/src/bin/acl.rs
awilliams-fastly: Yes, that's also how I was thinking the sequencing would be. 

This feature guard can be removed once the new `fastly` crate is published (the order of the previous steps shouldn't matter).

lib/src/config/acl.rs
awilliams-fastly: I choose this structure to match how other similar sections (e.g [`backends`](https://github.com/fastly/Viceroy/blob/6aae133d0a75cf423064b07c40ccce8830ee7e98/lib/src/config/backends.rs#L31), [`device_detection`](https://github.com/fastly/Viceroy/blob/6aae133d0a75cf423064b07c40ccce8830ee7e98/lib/src/config/device_detection.rs#L38), [`dictionaries`](https://github.com/fastly/Viceroy/blob/6aae133d0a75cf423064b07c40ccce8830ee7e98/lib/src/config/dictionaries.rs#L99), ...). So not a very compelling reason.

I'll change it to whichever organization you think best.

test-fixtures/data/my-acl-1.json
awilliams-fastly: Yes, good idea. I've made that change in [c18e3b6](https://github.com/fastly/Viceroy/pull/438/commits/c18e3b6e8234c0117b1e64ffc3078ab7079706e8)

lib/src/acl.rs
awilliams-fastly: Yes, it's ignored (along with any other extraneous fields).

Thanks for pointing out the inconsistency with this test case and the integration test data. In [c18e3b6](https://github.com/fastly/Viceroy/pull/438/commits/c18e3b6e8234c0117b1e64ffc3078ab7079706e8), I've corrected that along with adding better comments to better document the JSON structure.

ulyssa (approved):

PR #439, opened 2024-10-30
Add new `lookup_wait_v2` to fix generation parsing bug

computermouth:
Catching up to a bug fix in the upcoming SDK. Check PR 4427 in the other repository for more info.

Generations from the upstream are actually a u64. Parsing in 0.11.0 SDK and matching hostcalls always fails to 0.


ulyssa (commented):
This looks good, but I think you might need to run `make adapter` to update the `.wasm` for the new hostcall it's adapting.

ulyssa (dismissed):

ulyssa (approved):

PR #440, opened 2024-11-07
Add support for the on_behalf_of endpoints

acw:
These are a privileged set of operations, which we don't actually support in Viceroy. However, we want to make sure that these endpoints exist so that newer SDK builds don't run into any problems.

### DIFF
--- a/crates/adapter/src/fastly/core.rs
+++ b/crates/adapter/src/fastly/core.rs
@@ -1854,6 +1854,11 @@ pub mod fastly_http_req {
 
         res
     }
+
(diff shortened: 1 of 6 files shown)


ulyssa (commented):
This all looks good to me. If you run `make adapter` to update the adapter wasm to include the new hostcall, then I can approve the PR.

I'm going to also put up a PR to stub out the shielding hostcalls with unsupported errors, too, so that we can include those changes in the SDK.

ulyssa (approved):

PR #441, opened 2024-11-11
Add stubs for shielding support

ulyssa:
This adds stubs that return `FastlyStatus::UNSUPPORTED` for the two shielding hostcalls. I've set them up to do some minimal validation of their input before returning an error so that it can hopefully be useful for testing the JS and Go SDKs.


acw:
Subsumed in #455.

PR #445, opened 2024-12-05
Update Fastly crate from `0.10.4` to `0.11.1`

awilliams-fastly:
This updates the test fixtures to use the newly released `0.11.1` Fastly crate.

When `fastly::acl` hostcalls were added in #438, the corresponding SDK support was not yet released. So #438 added tests using the still-unreleased-SDK, but put that code behind a compile-time gate ([here](https://github.com/fastly/Viceroy/pull/438#discussion_r1823157613)).

This change removes that compile-time gate in `test-fixtures/src/bin/acl.rs`, effectively changing these tests from a no-op to actually running the desired test cases.

A few other text-fixtures stopped compiling with the updated SDK, which I've tried to address as well:

- `test-fixtures/src/bin/inspect.rs` cc/ @ulyssa

- `test-fixtures/src/bin/kv_store.rs` cc/ @joeshaw

The following hostcalls were missing from the component/adapter, and so were added:
- `fastly_http_req#send_v3`
- `fastly_http_req#send_async_v2`
- `fastly_http_cache` (entire module)


computermouth (approved):
KV Store tests look good to me :+1:

ulyssa (approved):

ulyssa:
It looks like the tests are failing because there isn't support for the `fastly_http_cache` module in the adapter.

ulyssa (approved):

ulyssa:
It looks like the macos-12 image is now completely unsupported (actions/runner-images#10721), so I've removed it from the required checks.

PR #446, opened 2025-01-17
extend `ulyssa/shielding-stubs` by removing the Unsupported error, and returning _something_.

lewiscowper:
This allows for calls that use the shielding APIs, as they are, not to crash. I've no idea how correct, useful, or otherwise relevant the output is, but I went from my tests failing to passing after making these adjustments locally.


lib/src/wiggle_abi/shielding.rs
```diff
@@ -7,16 +7,29 @@ impl fastly_shielding::FastlyShielding for Session {
...
         _out_buffer_max_len: u32,
     ) -> Result<u32, Error> {
         // Validate the input name and then return the unsupported error.
         let name_bytes = memory.to_vec(name.as_bytes())?;
-        let _name = String::from_utf8(name_bytes).map_err(|_| Error::InvalidArgument)?;
+        let name = String::from_utf8(name_bytes).map_err(|_| Error::InvalidArgument)?;
+
+        let mut shield_config = name.split('-');
+
+        let first = shield_config.next().expect("first is required");
```
lewiscowper, line 19: I leave it up to your discretion to rename these as makes sense, I'm not 100% sure what the blocks refer to.

lib/src/wiggle_abi/shielding.rs
```diff
@@ -0,0 +1,64 @@
...
+    ) -> Result<u32, Error> {
+        // Validate the input name and then return the unsupported error.
+        let name_bytes = memory.to_vec(name.as_bytes())?;
+        let name = String::from_utf8(name_bytes).map_err(|_| Error::InvalidArgument)?;
+
+        let mut shield_config = name.split('-');
+
+        let first = shield_config.next().expect("first is required");
+        let second = shield_config.next().expect("second is required");
+        let _third = shield_config.next().expect("third is required");
```
lewiscowper, line 21: It also seems that three pieces aren't always present in Shield names. So this is wrong.

PR #447, opened 2025-01-28
Update KV store key naming restrictions

joeshaw:
Conform to the restrictions in
https://docs.fastly.com/en/guides/working-with-kv-stores#limitations-and-considerations

Partially fixes https://github.com/fastly/cli/issues/1383

### DIFF
--- a/lib/src/object_store.rs
+++ b/lib/src/object_store.rs
@@ -439,6 +439,8 @@ impl From<&KvStoreError> for FastlyStatus {
 ///   * Keys cannot contain Carriage Return or Line Feed characters.
 ///   * Keys cannot start with `.well-known/acme-challenge/`.
 ///   * Keys cannot be named `.` or `..`.
+///   * Keys cannot use Unicode characters 0 through 32, 65534 and 65535 as
+///     single-character key names.  (0x0 through 0x20, 0xFFFE and 0xFFFF)
 fn is_valid_key(key: &str) -> Result<(), KeyValidationError> {
     let len = key.as_bytes().len();
     if len < 1 {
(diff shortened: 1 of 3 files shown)


lib/src/object_store.rs
```diff
@@ -439,6 +439,8 @@ impl From<&KvStoreError> for FastlyStatus {
 ///   * Keys cannot contain Carriage Return or Line Feed characters.
 ///   * Keys cannot start with `.well-known/acme-challenge/`.
 ///   * Keys cannot be named `.` or `..`.
+///   * Keys cannot use Unicode characters 0 through 32, 65534 and 65535 as
```
cceckman-at-fastly, line 442: I still see two differences from the documentation:

1. The documentation prohibits `..` as a prefix:

> Keys cannot start with [...] `..`.

But here it's (only) enforced as a full-match. `../privkey.pem` is permitted by this filter but not by production.

2. This says "keys cannot be named `.`", but that restriction isn't in the public docs.

I don't know which restrictions match the production service, but we should probably align everything to it.

cceckman-at-fastly (approved):
Approving, as this looks like an improvement even with the below comment :)

lib/src/object_store.rs
joeshaw: Good points.  I've updated the implementation to better match the production service in the key restrictions.

joeshaw:
I added the additional restrictions in e05425b and rebased on the current main branch after #448 was merged.  @cceckman-at-fastly PTAL.

cceckman-at-fastly (approved):

PR #448, opened 2025-02-21
Buildfix: get CI working

### LINKED ISSUE #444: Undeclared dependence on newer Rust features
v0.12.2 relies on `#is_non_or` from Rust 1.82.0 and `#is_bits` stabilization from Rust 1.83.0 (if I'm not mistaken from looking at changelogs).

I previously had Rust 1.78.0 on my machine and attempting to install resulted in:

```
error[E0599]: no method named `is_none_or` found for enum `std::option::Option` in the current scope
  --> /home/geemus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/viceroy-lib-0.12.2/src/acl.rs:61:24
   |
61 |                 if acc.is_none_or(|prev_match: &Entry| mask >= prev_match.prefix.mask) {
   |                        ^^^^^^^^^^
   |
help: there is a method `is_none` with a similar name, but with different arguments
  --> /tmp/rust-20240518-13478-4dthw7/rustc-1.78.0-src/library/core/src/option.rs:651:5

error[E0658]: use of unstable library feature 'ip_bits'
  --> /home/geemus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/viceroy-lib-0.12.2/src/acl.rs:92:32
   |
92 |                     IpAddr::V4(Ipv4Addr::from_bits(v4.to_bits() & bit_mask)),
   |                                ^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #113744 <https://github.com/rust-lang/rust/issues/113744> for more information

error[E0658]: use of unstable library feature 'ip_bits'
  --> /home/geemus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/viceroy-lib-0.12.2/src/acl.rs:92:55
   |
92 |                     IpAddr::V4(Ipv4Addr::from_bits(v4.to_bits() & bit_mask)),
   |                                                       ^^^^^^^
   |
   = note: see issue #113744 <https://github.com/rust-lang/rust/issues/113744> for more information

error[E0658]: use of unstable library feature 'ip_bits'
   --> /home/geemus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/viceroy-lib-0.12.2/src/acl.rs:100:32
    |
100 |                     IpAddr::V6(Ipv6Addr::from_bits(v6.to_bits() & bit_mask)),
    |                                ^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #113744 <https://github.com/rust-lang/rust/issues/113744> for more information

error[E0658]: use of unstable library feature 'ip_bits'
   --> /home/geemus/.cargo/registry/src/index.crates.io-6f17d22bba15001f/viceroy-lib-0.12.2/src/acl.rs:100:55
    |
100 |                     IpAddr::V6(Ipv6Addr::from_bits(v6.to_bits() & bit_mask)),
    |                                                       ^^^^^^^
    |
    = note: see issue #113744 <https://github.com/rust-lang/rust/issues/113744> for more information
```

We also have github actions that use viceroy which were failing with the `is_non_or` problem. It attempts to pull the latest viceroy, but is relying on whatever version of rust is available, which will be 1.83.0 it looks like from github-actions/runner-images, but isn't there yet. We will probably also fix this to lock the viceroy version so it advances when we tell it to, rather than always tracking latest.

Updating to Rust 1.83.0 fixed both kinds of issues, which was easy enough on my machine, but is more complicated in github actions.

Anyway, I wanted to call this out as it would be nice if the cargo manifest explicitly specified the minimum expected Rust version (or else the code was more backwards compatible).

cceckman-at-fastly:
- `rust-toolchain` says we should use `stable` to build... but the two most recent stable releases (1.84 and 1.85) use `wasm32-wasip1`, not `wasm32-wasi`.

  We should either keep `rust-toolchain` at 1.83, or update per the upgrade (this change), so that the code is compatible with the stated toolchain.
  
  `stable` seems to cause some additional issues on Windows, so -- 1.83 it is.
  
  Still, update to `wasm32-wasip1` as the target name for tests. This doesn't affect anything other than Viceroy's own testing -- I used this Viceroy build to `fastly compute serve` a project built as `wasm32-wasi` on 1.83.
  
- The `adapter` package expected a `std` feature in the `witx`-generated code, but didn't declare the feature. I guess this wasn't an error in previous versions? Since the package relies on `Vec` anyway, drop the `std_feature` declaration.

- Include minimum supported Rust versions. Fixes #444 .


cceckman-at-fastly:
Amusingly, CI failed on the first go for this exact reason: the CI workers pulled `stable` but couldn't install `wasm32-wasi`.

joeshaw (approved):
👍

PR #449, opened 2025-03-04
Include FASTLY_IS_STAGING environment variable

cceckman-at-fastly:
See https://docs.fastly.com/en/guides/working-with-staging for more on staging.

This environment variable [indicates](https://docs.fastly.com/en/guides/working-with-staging#compute) the request is running in the staging environment. Viceroy is not "the staging environment", so we return 0. (Admittedly, it's also not running in production...)

Relevant to #336.

### DIFF
--- a/test-fixtures/src/bin/env-vars.rs
+++ b/test-fixtures/src/bin/env-vars.rs
@@ -31,4 +31,7 @@ fn main() {
 
     let host_name = env::var("FASTLY_HOSTNAME").expect("host name available");
     assert_eq!(host_name, "localhost");
+
+    let is_staging = env::var("FASTLY_IS_STAGING").expect("staging variable set");
+    assert!(is_staging == "0" || is_staging == "1");
 }
--- a/test-fixtures/src/bin/env-vars.rs
+++ b/test-fixtures/src/bin/env-vars.rs
@@ -33,5 +33,5 @@ fn main() {
     assert_eq!(host_name, "localhost");
 
     let is_staging = env::var("FASTLY_IS_STAGING").expect("staging variable set");
-    assert!(is_staging == "0" || is_staging == "1");
+    assert!(is_staging == "0");
 }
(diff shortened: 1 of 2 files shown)


test-fixtures/src/bin/env-vars.rs
```diff
@@ -31,4 +31,7 @@ fn main() {
 
     let host_name = env::var("FASTLY_HOSTNAME").expect("host name available");
     assert_eq!(host_name, "localhost");
+
+    let is_staging = env::var("FASTLY_IS_STAGING").expect("staging variable set");
+    assert!(is_staging == "0" || is_staging == "1");
```
franfastly, line 36: Is this test checking the stub values set by default? I'm not sure when `FASTLY_IS_STAGING` could be `1` in this case.

cceckman-at-fastly: Yeah - this is me deviating from "what the test does" to "what I think it should do".

I think a directed test should accept any correct implementation. e.g.: I don't think the test should reject an implementation that allows for arbitrary variables (#336). That would make it a [change-detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).

Which - it is today, this would fail if we implemented #336. Except for the new part of the test, which says *only* "the input must be from the valid range". (Since we documented `FASTLY_IS_STAGING=0` to mean "production", 0 and 1 are both ~equally wrong.)

Updated to test that it's 0 for now, to keep it consistent; but I'll make another PR to make my argument.

cceckman-at-fastly: Put the directional change in #450 .

franfastly: Thank you for the clarification. I agree that there is value in checking valid ranges, maybe something we could do on runtime.

franfastly (approved):
Thank you! ✨

PR #450, opened 2025-03-04
Accept all valid environment variables

cceckman-at-fastly:
Pursuant to #336.

If we were to implement #336 today, this test would fail off the bat:
not because the implementation is incorrect, but because the test
does not verify correct behavior, only the current behavior.
That is, it is a change-detector test [1].

A directed test should admit any correct implementation.

To that end, refactor so that this tests presence and (where applicable)
structure of environment variables, but does not validate that they have
particualr values. (Except in the case of `FASTLY_HOSTNAME`,
which has one correct value for Viceroy.)

[1]: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html


cee-dub (approved):
Yes!!

cee-dub (approved):
No problem. I have wanted environment variable support for a long time, and this brings me hope.

PR #451, opened 2025-03-05
Remove ubuntu-20.04 from CI

cceckman-at-fastly:
Per https://github.com/actions/runner-images/issues/11101, Github is turning down 20.04 actions, and is issuing brownouts that cause CI failures. (I saw it fail the 20.04 matrix steps and the package-check step on my PRs.)

To keep the CI flowing, remove 20.04 from the matrices, and upgrade package-check to 24.04.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,15 +10,12 @@ jobs:
     strategy:
       matrix:
         rust-toolchain: ["1.83"]
-        os: [ubuntu-20.04, macos-14, windows-2022]
+        os: [macos-14, windows-2022]
         arch: [amd64, arm64]
         exclude:
           - os: windows-2022
             arch: arm64
         include:
-          - os: ubuntu-20.04
-            name: linux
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -10,15 +10,12 @@ jobs:
     strategy:
       matrix:
         rust-toolchain: ["1.83"]
-        os: [ubuntu-20.04, macos-14, windows-2022]
+        os: [macos-14, windows-2022]
```
kpfleming, line 13: Should this be a change from 20.04 to 24.04 instead of removal of Ubuntu entirely?

cceckman-at-fastly: Absolutely! Overzealous deletes on my part

kpfleming (approved):

cceckman-at-fastly:
For the further record: I also removed the branch protection rules that required results from 20.04, as those were preventing this from going in :smile:

PR #452, opened 2025-03-05
Add CollectingBody for cache use cases

cceckman-at-fastly:
`CollectingBody` reads from a `Body` and:

- Stores the data and trailers written, for future delivery in cache
  lookups
- Offers a `read() -> Body` endpoint. The read can proceed concurrently
  with writes, or after the write has completed.

### DIFF
--- a/lib/src/collecting_body.rs
+++ b/lib/src/collecting_body.rs
@@ -1,0 +1,340 @@
+//! Provides `CollectingBody`, a body that can be concurrently written-to (via a `StreamingBody`
+//! handle) and read-from (via multiple `Body`) handles.
+
+// TODO: cceckman-at-fastly: Temporary, I promise.
+#![allow(unused)]
+
+use bytes::Bytes;
+use http::HeaderMap;
+use http_body::Body as HttpBody;
+use tokio::sync::watch;
+
+use crate::{
+    body::{Body, Chunk},
+    error,
+    streaming_body::StreamingBody,
+    Error,
+};
+
+/// CollectingBody is a body for caching and request collapsing.
+/// It allows writing a body while concurrently\* reading it, from multiple readers.
+#[derive(Debug)]
+pub struct CollectingBody {
+    inner: watch::Receiver<CollectedBody>,
+}
+
+/// Body consisting strictly of Byte chunks, with concurrent readying and writing.
+///
+/// CollectingBody primarily exists to implement the cache APIs. Cache APIs allow three things to
+/// happen concurrently:
+/// - Streaming in a body, from an `insert`/`replace`-style request, implemented via StreamingBody
+/// - Storage of that body in the cache
+/// - Streaming out of the body, from a `lookup`/`TransactionInsertBuilder:execute_and_stream_back`
+///   request, from the same or a different session
+///
+/// CollectingBody provides a place for this to happen. It accepts a `Body` as a source of data,
+/// e.g. one from a `StreamingBody` or an origin response; stores the data for future retrieval;
+/// and can return new `Body`s from `::read` that produce the full content.
+///
+/// A CollectingBody is finished when either error.is_some() or trailers.is_some().
+/// Until then, more chunks may be added!
+#[derive(Default, Debug)]
+struct CollectedBody {
+    // We preserve "multiple Byte chunks" here to avoid additional copies-
+    // the Byte type uses a shared buffer under the hood.
+    // TODO: cceckman-at-fastly: consider SmallVec
+    chunks: Vec<Bytes>,
+    error: Option<Error>,
+    trailers: Option<HeaderMap>,
+}
+
+impl CollectingBody {
+    /// Create a new CollectingBody that stores & streams from the provided Body.
+    ///
+    /// Writes to the StreamingBody are collected, and propagated to all readers of this
+    /// CollectingBody.
+    // TODO: Expected length?
+    pub fn new(from: Body) -> CollectingBody {
+        let (tx, rx) = watch::channel(CollectedBody::default());
+        let body = CollectingBody { inner: rx };
+        tokio::task::spawn(Self::tee(from, tx));
+        body
+    }
+
+    /// "tee" a single Body to the watch channel.
+    ///
+    /// This is the worker thread behind a CollectingBody. It reads the data from the provided body
+    /// into a `tokio::sync::watch` channel, which (a) accumulates the body + trailers and (b)
+    /// notifies any subscribed readers of the updates. The readers can safely miss updates or
+    /// start late, as they always can eventually read the state.
+    async fn tee(mut rx: Body, tx: watch::Sender<CollectedBody>) {
+        // IMPORTANT IMPLEMENTATION NOTE:
+        //
+        // Make sure every path out of this function results in either state.errors.is_some() or
+        // state.trailers.is_some().
+
+        // Read data first:
+        while let Some(chunk) = rx.data().await {
+            match chunk {
+                Ok(data) => {
+                    tx.send_modify(move |state| {
+                        state.chunks.push(data);
+                    });
+                }
+                Err(Error::Again) => continue,
+                Err(e) => {
+                    tx.send_modify(move |state| {
+                        state.error = Some(e);
+                    });
+                    return;
+                }
+            }
+        }
+
+        // Then wait for trailers (if any) to be present:
+        let trailers = rx.trailers().await;
+        tx.send_modify(move |state| match trailers {
+            Ok(v) => state.trailers = Some(v.unwrap_or_default()),
+            Err(e) => state.error = Some(e),
+        });
+    }
+
+    /// Get a new read handle to this body.
+    pub fn read(&self) -> Result<Body, error::Error> {
+        let mut upstream = self.inner.clone();
+        let (mut tx, rx) = StreamingBody::new();
+        tokio::task::spawn(async move {
+            let mut next_chunk = 0;
+
+            loop {
+                // We can't hold current_value across an await point,
+                // and borrowck doesn't undersand drop() enough for us to use it in a branch.
+                // Queue up our work, then do it.
+                let send_chunks: Vec<Bytes>;
+                let trailers: Option<HeaderMap>;
+                let error: Option<String>;
+                if let Ok(current_value) = upstream
+                    .wait_for(|v| {
+                        v.trailers.is_some() || v.error.is_some() || v.chunks.len() > next_chunk
+                    })
+                    .await
+                {
+                    send_chunks = (&current_value.chunks[next_chunk..])
+                        .iter()
+                        .cloned()
+                        .collect();
+                    trailers = current_value.trailers.clone();
+                    // The Error type is not Clone, so we can't get at it directly.
+                    // But we can get a stringification of it.
+                    error = current_value.error.as_ref().map(|e| e.to_string());
+                } else {
+                    // Channel is closed.
+                    //
+                    // This should only happen if the sender has hung up, i.e. if the object has
+                    // been evicted from the cache *and* the writer is done. In which case:
+                    // - If the writer shut down cleanly, great, we've already shut down cleanly
+                    //      too.
+                    // - If the writer didn't, we don't either.
+                    return;
+                }
+
+                // OK, we have some data. Forward it. In order,
+                // Good data:
+                for chunk in send_chunks {
+                    if tx.send_chunk(chunk).await.is_err() {
+                        // Reader hung up; we don't care any more.
+                        return;
+                    }
+                    next_chunk += 1;
+                }
+                // Finished data:
+                if let Some(trailers) = trailers {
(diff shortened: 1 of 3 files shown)


lib/src/collecting_body.rs
```diff
@@ -0,0 +1,340 @@
...
+struct CollectedBody {
+    // We preserve "multiple Byte chunks" here to avoid additional copies-
+    // the Byte type uses a shared buffer under the hood.
+    // TODO: cceckman-at-fastly: consider SmallVec
+    chunks: Vec<Bytes>,
+    error: Option<Error>,
+    trailers: Option<HeaderMap>,
+}
+
+impl CollectingBody {
```
drbrain, line 51: Since `CollectingBody` and `CollectedBody` are very similar names can you swap the `struct` ordering so the `struct` and `impl` for `CollectingBody` are adjacent? It gave me a moment of confusion when reading the impl.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,340 @@
...
+    ///
+    /// This is the worker thread behind a CollectingBody. It reads the data from the provided body
+    /// into a `tokio::sync::watch` channel, which (a) accumulates the body + trailers and (b)
+    /// notifies any subscribed readers of the updates. The readers can safely miss updates or
+    /// start late, as they always can eventually read the state.
+    async fn tee(mut rx: Body, tx: watch::Sender<CollectedBody>) {
+        // IMPORTANT IMPLEMENTATION NOTE:
+        //
+        // Make sure every path out of this function results in either state.errors.is_some() or
+        // state.trailers.is_some().
```
drbrain, line 74: In CacheD for body handling we have an implementation similar to this, but use an enum to indicate the state of the stream. Something like:

```
#[derive(Default)]
enum CollectState {
    #[default]
    Streaming(CollectingBody),
    Complete(CollectedBody),
    Error(…),
}

pub struct CollectingBody {
    inner: watch::Receiver<CollectState>,
}
```

Where your `CollectingBody::error` moves to `CollectState::Error` and `CollectingBody::trailers` moves to a new `CollectedBody` struct.

It's a bit more work, but the compiler will reduce the possibility of an implementation error.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,340 @@
...
+            loop {
+                // We can't hold current_value across an await point,
+                // and borrowck doesn't undersand drop() enough for us to use it in a branch.
+                // Queue up our work, then do it.
+                let send_chunks: Vec<Bytes>;
+                let trailers: Option<HeaderMap>;
+                let error: Option<String>;
+                if let Ok(current_value) = upstream
+                    .wait_for(|v| {
+                        v.trailers.is_some() || v.error.is_some() || v.chunks.len() > next_chunk
```
drbrain, line 118: This can always be true, so you can switch to `.changed().await`, if you only call `tx.send_modify()` to push new data when the chunk of data has at least one byte up on L80.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,340 @@
...
+                    //
+                    // This should only happen if the sender has hung up, i.e. if the object has
+                    // been evicted from the cache *and* the writer is done. In which case:
+                    // - If the writer shut down cleanly, great, we've already shut down cleanly
+                    //      too.
+                    // - If the writer didn't, we don't either.
+                    return;
+                }
+
+                // OK, we have some data. Forward it. In order,
```
drbrain, line 141: The state pattern described above also means these comments become a `match`

lib/src/collecting_body.rs
drbrain: … I guess it already is always true, but you could collect zero-size chunks which only advance `next_chunk`

lib/src/collecting_body.rs
cceckman-at-fastly: Done.

I'm not sure the compiler is helping much in this case so much as the runtime; the need to deconstruct means it's more "panic if it isn't in the expected state". Still, sure.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+                // lock at the end of the scope.
+                let (send_chunks, trailers) = {
+                    // Acquire the read lock:
+                    let current_value = upstream.borrow_and_update();
+                    let send_chunks: Vec<Bytes> = current_value.chunks().skip(next_chunk).collect();
+                    let trailers = current_value.trailers().cloned();
+                    if current_value.is_error() {
+                        // To trigger a guest error, it is sufficient to
+                        // TODO: cceckman-at-fastly: Do we need to do something with the error message?
+                        // To trigger a guest error, it appears sufficient to not .finish() the StreamingBody.
```
aturon, line 134: Needs a bit of tidying.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+                        return;
+                    }
+                    (send_chunks, trailers)
+                };
+
+                // Good data:
```
aturon, line 140: What does "good data" mean here?

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+    Streaming(Vec<Bytes>),
+    Complete {
+        body: Vec<Bytes>,
+        trailers: HeaderMap,
+    },
+    Error(Error),
+}
+
+impl CollectingBodyInner {
+    fn chunks(&self) -> impl Iterator<Item = Bytes> + use<'_> {
```
aturon, line 181: Might be simpler to just return a `&[Bytes]` here?

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+        static EMPTY_VEC: Vec<Bytes> = Vec::new();
+        match self {
+            CollectingBodyInner::Streaming(body) | CollectingBodyInner::Complete { body, .. } => {
+                body.iter().cloned()
+            }
+            _ => EMPTY_VEC.iter().cloned(),
+        }
+    }
+    fn trailers(&self) -> Option<&HeaderMap> {
```
aturon, line 190: The use of `Option` here to signal "completion" is a bit subtle. Might be nice to document that.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+        let reader = {
+            let data = Arc::clone(&data);
+            tokio::task::spawn(async move {
+                let read = collect.read().unwrap();
+                let bytes = read.read_into_vec().await.unwrap();
+                assert_eq!(&bytes, &*data);
+            })
+        };
+        // Try to catch the reader up:
+        tokio::task::yield_now().await;
```
aturon, line 298: A load-bearing yield makes me nervous. Could this and below be done via one-shot blocking instead?

aturon (changes requested):
Clean and clever! Just a few nits and notes.

lib/src/collecting_body.rs
cceckman-at-fastly: Done in 3537476

lib/src/collecting_body.rs
cceckman-at-fastly: Done in 3537476

lib/src/collecting_body.rs
cceckman-at-fastly: "If this is nonempty, it's data to forward" - revised the comment in 3537476.

lib/src/collecting_body.rs
cceckman-at-fastly: Done in 3537476

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub fn new(from: Body) -> CollectingBody {
+        let (tx, rx) = watch::channel(CollectingBodyInner::default());
+        let body = CollectingBody { inner: rx };
+        tokio::task::spawn(Self::tee(from, tx));
+        body
+    }
+
+    /// "tee" a single Body to the watch channel.
+    ///
+    /// This is the worker thread behind a CollectingBody. It reads the data from the provided body
```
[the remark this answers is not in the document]
cceckman-at-fastly: Thanks -- took a couple tries to figure out how to not re-implement `Body`.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,367 @@
...
+
+use crate::{
+    body::{Body, Chunk},
+    error,
+    streaming_body::StreamingBody,
+    Error,
+};
+
+/// CollectingBody is a body for caching and request collapsing.
+/// It allows writing a body while concurrently\* reading it, from multiple readers.
```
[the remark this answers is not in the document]
cceckman-at-fastly: Done in 3537476

lib/src/collecting_body.rs
cceckman-at-fastly: Done in 9287ec6.

I'm not sure if any of these were load-bearing in the sense of correctness-of-the-program, but yes, the intended test sequence depended on them, which isn't great.

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,387 @@
...
+        body: Vec<Bytes>,
+        trailers: HeaderMap,
+    },
+    Error(Error),
+}
+
+impl CollectingBodyInner {
+    /// The chunks currently available.
+    fn chunks(&self) -> &[Bytes] {
+        static EMPTY_VEC: Vec<Bytes> = Vec::new();
```
drbrain, line 185: Unused?

```suggestion
```

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,387 @@
...
+#[derive(Debug)]
+pub struct CollectingBody {
+    inner: watch::Receiver<CollectingBodyInner>,
+}
+
+impl Default for CollectingBodyInner {
+    fn default() -> Self {
+        CollectingBodyInner::Streaming(Vec::default())
+    }
+}
```
drbrain, line 41: I think this can become a `derive(Default)` by adding `#[default]` above the `Streaming` variant of the enum, [docs here](https://doc.rust-lang.org/stable/std/default/trait.Default.html#enums)

lib/src/collecting_body.rs
cceckman-at-fastly: Thanks! Done in b2a4195

lib/src/collecting_body.rs
cceckman-at-fastly: Thanks, yep! Done in b2a4195

cceckman-at-fastly: Ah, right -- I forgot why I had it that way:

```
error: the `#[default]` attribute may only be used on unit enum variants
   --> /home/runner/work/Viceroy/Viceroy/lib/src/collecting_body.rs:169:5
    |
169 |     Streaming(Vec<Bytes>),
    |     ^^^^^^^^^
    |
    = help: consider a manual implementation of `Default`
```

lib/src/collecting_body.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /// Get a new read handle to this body.
+    pub fn read(&self) -> Result<Body, error::Error> {
+        let mut upstream = self.inner.clone();
+        let (mut tx, rx) = StreamingBody::new();
+        tokio::task::spawn(async move {
+            let mut next_chunk = 0;
+
+            while upstream.changed().await.is_ok() {
```
drbrain, line 113: Is it possible to crate a `CollectingBody` after the writer has written the entire document? The current tests don't seem to cover that scenario as the `CollectingBody` is created before the `StreamingBody` calls `finish()` on `tx`.

If it is possible to create one after the transmitter finishes I think this will either block forever or error because cloning a `watch::Receiver` doesn't seem to cause the current value to be unseen. From the [change notifications] section of `tokio::watch`:

> At creation, the initial value is considered seen. In other words, [Receiver::changed()](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed) will not return until a subsequent value is sent.

lib/src/collecting_body.rs
drbrain: 😢

aturon (commented):
Changes look good to me! Approved once @drbrain's concern is addressed.

lib/src/collecting_body.rs
cceckman-at-fastly: Nice catch!

> Is it possible to create a `CollectingBody` after the writer has written the entire body?

It _shouldn't_ be possible using the cache APIs -- I had considered having the `CollectingBody()` constructor return a `StreamingBody, CollectingBody)` pair -- but it's not impossible for later.

I agree with your read, though when I added a test it didn't have a problem- guessing that the tokio scheduler just wasn't tickled into quite the right interleaving, since it depends on when `tee` runs vs. everything else. Will write a patch with the test & a fix.

drbrain (approved):
My name is Eric Hodel and I approve this pull request.

aturon (approved):

PR #454, opened 2025-03-10
Partial implementation of cache API

cceckman-at-fastly:
The minimum possible implementation. Limitations:

- Only `insert` and `lookup`; no transactions
- Just key -> body; no TTL, eviction, etc.; no options

Since the insert/lookup options are not supported, the `insert` and `lookup` methods still return unavailable by default. They are enabled at runtime by `ENABLE_EXPERIMENTAL_CACHE_API=1` in the environment.


lib/src/cache.rs
```diff
@@ -66,3 +218,49 @@ impl CacheOverride {
...
+        #[test]
+        fn accept_valid_cache_key_len(l in 0usize..4096) {
+            let mut v : Vec<u8> = Vec::new();
+            v.resize(l, 0);
+            let _ = CacheKey::try_from(&v).unwrap();
+        }
+    }
+
+    proptest! {
+        #[test]
```
drbrain, line 247: Is `#[tokio::test]` not usable inside `proptest!`?

cceckman-at-fastly: Unfortunately, yeah; both of them are doing Things to the signatures of the relevant functions, and they aren't orthogonal. Added a comment.

lib/src/wiggle_abi/cache.rs
```diff
@@ -215,49 +265,57 @@ impl FastlyCache for Session {
...
         options: &types::CacheGetBodyOptions,
     ) -> Result<types::BodyHandle, Error> {
-        Err(Error::NotAvailable("Cache API primitives"))
+        // TODO: cceckman-at-fastly ; options
+        let entry = self.cache_entry_mut(handle).await?;
+
+        let Some(found) = entry.found() else {
+            return Err(Error::CacheError("key was not found in cache".to_owned()));
+        };
+        let body = found.body()?;
```
aturon, line 281: Note, CacheD has some restrictions around concurrent body streaming, which we (probably? eventually?) should match here. I don't recall the exact details off the top of my head.

lib/src/cache.rs
```diff
@@ -1,6 +1,158 @@
...
+use fastly_shared::FastlyStatus;
 use http::HeaderValue;
 
+mod store;
+
+use store::{CacheData, CacheKeyObjects};
+
+/// Primary cache key: an up-to-4KiB buffer.
+///
+// TODO: cceckman: use an inline-vec to make this cheaper to pass around
```
aturon, line 16: I'm curious about this TODO comment -- what do you have in mind here?

aturon (approved):
This looks great! I left a couple of notes/questions, but nothing that should block landing this PR. ✨

lib/src/wiggle_abi/cache.rs
cceckman-at-fastly: Per conversations with @drbrain , the restrictions are mostly implementation details in the internal-to-Fastly protocol. They show up here as:

- "`insert_and_stream_back` is a separate hostcall": separate "insert" then "lookup" is not guaranteed to get back the body you're streaming in, because another session may race to insert. `insert_and_stream_back` does some extra stuff under the hood to make sure it's consistent. The race with `insert` and `lookup` is true here as well; I haven't worked on `insert_and_stream_back` yet, but it should also be possible in this implementation to make it guarantee the right body (doing it in a single critical section).

- "The headers show up first" - which is also true in this implementation.

lib/src/cache.rs
cceckman-at-fastly: My thought was that we have special casing for 32B cache keys in Compute; something like `tinyvec`could avoid some allocation and indirection. That said - probably a premature optimization; 32B spills over a `tinyvec`-type structure anyway. Removed the comment.

lib/src/wiggle_abi/cache.rs
aturon: Ah sorry -- I meant the body-specific restrictions, listed [here](https://docs.rs/fastly/latest/fastly/cache/core/struct.Found.html#method.to_stream). We might want to enforce that in Viceroy, even though the implementation doesn't require it, just to help users test with this restriction.

lib/src/cache.rs
aturon: Got it, thanks!

lib/src/wiggle_abi/cache.rs
cceckman-at-fastly: Aha - yes, acknowledged. Thinking about ways to make this "not too invasive"; ultimately, the `Found` method needs to synchronize with the copying thread, so it can detect completion. (Could use a callback or similar, but that's not very async-style.)

So: a revision to `CollectingBody` where `.read()` returns a (future, body), perhaps. Will poke at it some more.

cceckman-at-fastly: I realized anything based on an "after all bytes are read" wouldn't be deterministic, and wouldn't map to something in the customer's control. So the constraint imposed (and tested) here: the `Body` from a `Found` must be `close()`d before getting another `Body`.

aturon (approved):
Thanks, the body handling looks great, and I appreciate the commentary as well! ✨

aturon (approved):

PR #456, opened 2025-03-18
Guest Profile sample period configuration support

posborne:
This change matches the configuration support for guest profling provided by wasmtime and is useful for integrators making use of guest profiling to be able to get a more fine/coarse grained view of the execution of their services (and also balance data generated, overhead, etc.).

In preparing to submit this change for review, I first saw that there is an alternative implementation of this with
https://github.com/fastly/Viceroy/pull/437, doing mostly the same things in a slightly different form.

### DIFF
--- a/cli/src/opts.rs
+++ b/cli/src/opts.rs
@@ -1,6 +1,8 @@
 //! Command line arguments.
 
-use viceroy_lib::config::UnknownImportBehavior;
+use std::time::Duration;
+
+use viceroy_lib::{config::UnknownImportBehavior, GuestProfileConfig};
 
 use {
     clap::{Args, Parser, Subcommand, ValueEnum},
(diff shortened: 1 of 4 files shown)


cli/src/opts.rs
```diff
@@ -314,6 +316,39 @@ fn check_module(s: &str) -> Result<PathBuf, Error> {
...
+/// Parse a string as a duration
+///
+/// This implementation is mostly borrowed from the wasmtime cli
+fn parse_profile_sample_duration(s: &str) -> Result<Duration, Error> {
+    // assume an integer without a unit specified is a number of seconds ...
+    if let Ok(val) = s.parse() {
+        return Ok(Duration::from_secs(val));
+    }
+
+    if let Some(num) = s.strip_suffix("s") {
```
cceckman-at-fastly, line 328: Nit: a little weird that we're intentionally taking the `num.parse()` `Err` path in the case of a sub-second suffix - it' be clearer IMO if `s` was the last case.  (It's correct as-written because of the fall-through, I just had to read it twice to understand how control would flow.)

cli/src/opts.rs
```diff
@@ -92,9 +94,12 @@ pub struct SharedArgs {
     ///
     /// The `guest` option can be additionally configured as:
     ///
-    ///     --profile=guest[,path]
+    ///     --profile=guest[[,path],sample]
```
cceckman-at-fastly, line 97:
```suggestion
    ///     --profile=guest[,path[,sample]]
```

`[]` indicate a layer of optionality, yes?

Based on the parser further down, you can have `path` alone (good, keeps backwards compatibility), or `path` and `sample`, but not `sample` without `path`.

cceckman-at-fastly (approved):

cli/src/opts.rs
posborne: Yeah, I definitely had this wrong.

cli/src/opts.rs
posborne: I agree, in this case I just pulled most of this impl was lifted from wasmtime as we're mirroring this cli flag: https://github.com/bytecodealliance/wasmtime/blob/main/crates/cli-flags/src/opt.rs#L414-L444

cceckman-at-fastly (approved):

PR #457, opened 2025-03-19
Age and vary_by

cceckman-at-fastly:
Implement two important metadata items:
- `max_age` and initial age, for computing staleness
- request/response headers and `vary_by`

And along with that... implement lookup according to HTTP's vary semantics, adding another layer to the internal data structures.

### DIFF
--- a/lib/src/cache/store.rs
+++ b/lib/src/cache/store.rs
@@ -68,7 +68,7 @@ impl ObjectMeta {
     }
 
     /// The vary rule associated with this request.
-    pub fn vary_rule(&self) -> Option<&VaryRule> {
+    pub fn vary_rule(&self) -> &VaryRule {
         &self.vary_rule
     }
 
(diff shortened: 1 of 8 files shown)


posborne (approved):
I looked over the changes here and they seem very reasonable to me.  There are a few TODOs left around, but those seem intentional rather than changes I believe you intended to get around to addressing with this changeset.

cceckman-at-fastly:
Some of the TODOs were actually stale- removed them.

I've also rebased on the `main` branch, since it was a squash merge; but ~the same content.

lib/src/cache/store.rs
```diff
@@ -1,59 +1,161 @@
...
+    user_metadata: Option<Bytes>,
+    sensitive_data: Option<bool>,
+    */
+}
+
+impl ObjectMeta {
+    /// Retrieve the current age of this object.
+    pub fn age(&self) -> Duration {
+        // Age in this cache, plus age upon insertion
+        Instant::now().duration_since(self.inserted) + self.initial_age
```
posborne, line 52: Same effect, but I'll usually write this as `self.inserted.elapsed() + self.initial_age`

posborne (approved):
Rebased updates look good to me.

lib/src/cache/store.rs
cceckman-at-fastly: Agreed, that's better - done in 79d16c0

(CI status wasn't getting reported, so I wanted to push a change anyway :) )

posborne (approved):

PR #459, opened 2025-03-26
Add support for "metadata" field for local_server KV Store entry configs in fastly.toml

harmony7:
This PR adds support for an optional `metadata` string field in `[local_server.kv_stores]` of the `fastly.toml` config file.

It's used something like this:
```toml
[[local_server.kv_stores.example_store]]
key = "entry"
data = "this is some data"
metadata = "this is some metadata"
```

Now it can be accessed, e.g., in a JavaScript application:
```javascript
const entry = await kv.get('entry');
console.log('metadata', entry.metadataText());    // this is some metadata
console.log('text', await entry.text());          // this is some data
```

As described on [KV Store Item - Metadata](https://www.fastly.com/documentation/reference/api/services/resources/kv-store-item/#metadata), `metadata` can only be a UTF-8 string up to 2000 characters, so I didn't add support for non-strings or for external files at this time.

I've also updated the `file` / `format` handling as well so you can specify:
```json
{
  "entry": {
    "data": "this is some data",
    "metadata": "this is some metadata"
  }
}
```

this is done in a backwards-compatible way, so that the following two are functionally identical:
```json
{
  "entry": "this is some data",
}
```
```json
{
  "entry": {
    "data": "this is some data",
    "metadata": null
  }
}
```

### DIFF
--- a/lib/src/config/object_store.rs
+++ b/lib/src/config/object_store.rs
@@ -230,8 +230,8 @@ fn read_json_contents(
                 data: s,
                 metadata: None,
             },
-            serde_json::Value::Object(mut obj) => {
-                let data = obj.remove("data").ok_or_else(|| {
+            serde_json::Value::Object(obj) => {
+                let data = obj.get("data").ok_or_else(|| {
                     ObjectStoreConfigError::FileValueWrongFormat { key: key.clone() }
                 })?;
 
@@ -242,7 +242,7 @@ fn read_json_contents(
                     })?
                     .to_string();
 
-                let metadata = match obj.remove("metadata") {
+                let metadata = match obj.get("metadata") {
                     Some(val) => Some(
                         val.as_str()
                             .ok_or_else(|| ObjectStoreConfigError::FileValueWrongFormat {
--- a/lib/src/config/object_store.rs
+++ b/lib/src/config/object_store.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use toml::{toml, Value};
+use toml::Value;
 use {
     crate::{
         error::{FastlyConfigError, ObjectStoreConfigError},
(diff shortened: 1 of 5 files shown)


joeshaw (dismissed):
Looks great, thank you! 👍

lib/src/config/object_store.rs
```diff
@@ -195,14 +220,45 @@ fn read_json_contents(file: &Path) -> Result<HashMap<String, String>, ObjectStor
...
-            .as_str()
-            .ok_or_else(|| ObjectStoreConfigError::FileValueWrongFormat { key: key.clone() })?
-            .to_owned();
-        contents.insert(key, value);
+        let entry = match value {
+            serde_json::Value::String(s) => ObjectStoreEntry {
+                data: s,
+                metadata: None,
+            },
+            serde_json::Value::Object(mut obj) => {
```
kpfleming, line 233:
```suggestion
            serde_json::Value::Object(obj) => {
```
`obj` does not escape this block, so there is no reason to modify it, and thus it does not need to be mutable.

lib/src/config/object_store.rs
```diff
@@ -195,14 +220,45 @@ fn read_json_contents(file: &Path) -> Result<HashMap<String, String>, ObjectStor
...
+                })?;
+
+                let data = data
+                    .as_str()
+                    .ok_or_else(|| ObjectStoreConfigError::FileValueWrongFormat {
+                        key: key.clone(),
+                    })?
+                    .to_string();
+
+                let metadata = match obj.remove("metadata") {
```
kpfleming, line 245:
```suggestion
                let metadata = match obj.get("metadata") {
```

kpfleming (approved):
This looks great! I made some small suggestions to remove an unnecessary `mut`.

harmony7:
@kpfleming thank you for the discussion on get/remove and the mut.  I've applied your suggestions, and I'll merge this =)

harmony7:
evidently after I committed the suggestions I need a review again

joeshaw (approved):

PR #463, opened 2025-03-29
Add support for "file" entries in KV Stores defined in files

harmony7:
This is a followup to #459 that enables KV Stores defined using JSON files to have their values defined using paths to files.

In the same way that an entry can be written using TOML:
```toml
[[local_server.kv_stores.example_store]]
key = "entry"
file = "./path/to/file"
metadata = "this is some metadata"
```

It can now be written in JSON:
```json
{
  "entry": {
    "file": "./path/to/file",
    "metadata": "this is some metadata"
  }
}
```

`data` and `file` are mutually exclusive; it's an error if neither or both are specified.

This is useful especially for KV Store values that are not strings.


joeshaw (dismissed):
👍

harmony7:
I had to push a followup so please re-review.  thank you!

kpfleming (dismissed):

harmony7:
Thanks @kpfleming I accepted your suggestions and now I need another review.

kpfleming (approved):

harmony7:
I had to make one more push to make the test pass.
Sorry to trouble you again @kpfleming.

kpfleming (approved):

PR #464, opened 2025-04-03
fix: don't panic when using some unsupported features

fgsch:
Use the correct error so we eventually surface None.

Fixes core and component model.


ulyssa:
@fgsch I somehow missed this draft PR when it was created. I've made similar updates in #499 for JA3, JA4, and other absent downstream metadata, so this should be fixed when that PR lands.

PR #466, opened 2025-04-07
Instrument hostcalls when Guest Profiling

posborne:
This change matches the behavior of the wasmtime cli by tying in call-hook instrumentation to record additional profile data about the stack frames for each hostcall entry/exit.

This can be very useful in understanding the time guest code is spent blocked on hostcalls (which won't be sampled by epoch interrupts).

The hostcall data shows up in the "Marker Chart" tab of profiler.firefox.com when used to view a guest profile.  Here's an example using a test case with some compute interlaced with a 25ms sleep.

image


harmony7:
I may need some time to get to reviewing this.

harmony7 (approved):
Looks good to me!

PR #471, opened 2025-04-22
Transactional cache API, missing some options

cceckman-at-fastly:
Start implementing the transactional side of the core cache API.

Transactions begin with a `transaction_lookup` or `transaction_lookup_async`. When the `CacheBusyHandle` is `await`ed, it either returns a result immediately; returns an obligation indicating "you should fulfill this"; or blocks, until the obligatee abandons or fulfills their obligation.

NB, the implementation of `cancel` here matches a recent change to Fastly's Compute Platform. Previously, `close(BusyHandle)` would block until the obligation was fulfilled or granted to the closer; now, `close` doesn't block. This shouldn't really matter to anyone other than making e.g. timeouts to origin fail in better ways.

### DIFF
--- a/lib/src/cache/store.rs
+++ b/lib/src/cache/store.rs
@@ -307,7 +307,7 @@ struct CacheValue {
 }
 
 /// An obligation to fetch & update the cache.
-struct Obligation {
+pub struct Obligation {
     object: Arc<CacheKeyObjects>,
     request_headers: HeaderMap,
     variant: Variant,
@@ -365,5 +365,46 @@ impl CacheData {
     /// Access to object's metadata
     pub(crate) fn get_meta(&self) -> &ObjectMeta {
         &self.meta
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{sync::Arc, time::Duration};
+
+    use http::HeaderMap;
+
+    use crate::{body::Body, cache::WriteOptions};
+
+    use super::CacheKeyObjects;
+
+    #[tokio::test]
+    async fn test_single_obligation() {
+        let ko = Arc::new(CacheKeyObjects::default());
+
+        let mut set = tokio::task::JoinSet::new();
+        for _ in 0..4 {
+            set.spawn({
+                let ko = Arc::clone(&ko);
+
+                async {
+                    let (found, obligation) = ko.transaction_get(&HeaderMap::default()).await;
+                    // Either have the obligation to fetch, or was blocked until the obligation
+                    // completed.
+                    assert!(found.is_some() != obligation.is_some());
+
+                    if let Some(o) = obligation {
+                        let b: Body = "hello".as_bytes().into();
+                        o.complete(WriteOptions::new(Duration::from_secs(100)), b);
+                    }
+                    if let Some(f) = found {
+                        let body = f.body.read().unwrap().read_into_string().await.unwrap();
+                        assert_eq!(&body, "hello");
+                    }
+                }
+            });
+        }
+        let _ = set.join_all().await;
+        assert!(ko.get(&HeaderMap::default()).is_some())
     }
 }
--- a/lib/src/cache/store.rs
+++ b/lib/src/cache/store.rs
@@ -194,6 +194,10 @@ impl CacheKeyObjects {
                     .iter()
                     .map(|v| v.variant(request_headers))
                     .collect();
+                // These are the existing cache entries that would be valid for this request,
+                // taking into account vary rules.
+                // They may be stale or not-yet-filled (i.e. obligation-only); let's see what we
+                // can get out of them.
                 let response_keyed_objects: Vec<_> = response_keys
                     .iter()
                     .filter_map(|v| key_objects.objects.get(v))
@@ -263,7 +267,8 @@ impl CacheKeyObjects {
     ///
     /// If a clear_obligation is provided, clear the "obligated" bit on that Variant in the same
     /// transaction (so there's only one wakeup). Note the clear_obligation variant may differ from
-    /// the variant inserted.
+    /// the variant inserted: we place the obligation marker based on the _existing_ Vary rules,
+    /// but we insert based on the Vary rule received in the response.
     pub fn insert(
         &self,
         request_headers: HeaderMap,
--- a/lib/src/cache/store.rs
+++ b/lib/src/cache/store.rs
@@ -109,123 +109,128 @@ impl CacheKeyObjects {
 
         for vary_rule in key_objects.vary_rules.iter() {
             let response_key = vary_rule.variant(request_headers);
-            if let Some(object) = key_objects.objects.get(&response_key) {
-                match &object.transactional {
-                    TransactionState::Present(v) => return Some(Arc::clone(v)),
-                    _ => continue,
-                }
+            if let Some(object) = key_objects
+                .objects
+                .get(&response_key)
+                .and_then(|v| v.present.clone())
+            {
+                return Some(object);
             }
         }
         None
     }
 
-    /// Perform a transactional lookup-or-obligate:
-    /// - If an appropriate response is already available, return that response (Ok)
-    /// - If no task is currently fetching an appropriate response, return an obligation to fetch
-    ///     (Err)
-    /// - Otherwise, await the completion of an appropriate response, possibly obligating ourselves
-    ///     to fetch in the future if no active task winds up fetching something appropriate.
-    pub async fn transaction_get_or_obligate(
+    /// Perform a transactional lookup.
+    /// Return a CacheData if existing data were found (even if stale),
+    /// and return an Obligaton if the data need to be freshened.
+    pub async fn transaction_get(
         self: Arc<Self>,
         request_headers: &HeaderMap,
-    ) -> Result<Arc<CacheData>, Obligation> {
+    ) -> (Option<Arc<CacheData>>, Option<Obligation>) {
         let mut sub = self.0.subscribe();
 
-        // We may be waked by inappropriate responses multiple times. Stay in the loop until we get
-        // an obligation or we get what we want.
         loop {
-            let key_objects = sub.borrow_and_update();
-            let mut present = Vec::new();
-            let mut waiting = 0;
+            let mut awaitable = false;
+            {
+                // The read-locked portion.
+                // Note, this returns only fresh responses.
+                let key_objects = sub.borrow_and_update();
 
-            let response_keys: Vec<_> = key_objects
-                .vary_rules
-                .iter()
-                .map(|v| v.variant(request_headers))
-                .collect();
-            for response_key in &response_keys {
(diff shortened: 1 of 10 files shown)


lib/src/cache/store.rs
```diff
@@ -99,98 +99,256 @@ impl From<WriteOptions> for ObjectMeta {
...
+            // If one of the entries above was within the SWR period *and* had an obligation,
+            // we could go ahead and return it without generating an obligation.
+            // However, we have to proceed to the write lock to produce an obligation.
+            // So, expect some change to the above control flow in the SWR implementation.
+
+            // If we found an acceptable in-progress request while under the read lock,
+            // we can await. The subscription ensures that, even though we're
+            // "waiting outside of the lock", we'll still see the updated version.
+            if awaitable {
+                let _ = sub.changed().await;
```
aturon, line 164: This `await` point is where the "busy handle" concept is meant to capture: it gives a way to tell that the lookup has become part of a collapsed request and is not (currently) the designated go-getter -- _without blocking on the completion of the transaction_. Then async I/O operations like `select` and `is_ready` can be used to examine the state of the transaction.

aturon: (Note for the record: we discussed this over video, and @cceckman-at-fastly is going to revise this PR to align with the intentions for busy handles)

cceckman-at-fastly: Acknowledged - we talked about this in our 1:1.

This needs to work by:
1. Before we return a `busy_handle`, either generate an obligation, or determine that there is already an obligatee
2. When waiting on a busy handle, resolve to "found" or "obligation" (which may happen if the previous obligatee has abandoned it in the mean time)

cceckman-at-fastly: I think 479c6643 takes care of this. Not a huge refactor, but a split between the "able to wait" and "not" cases. The updated `ready_and_pending` function shows that the first lookup for a key reliably comes back `!pending`.

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+    /// If ok_to_wait, this may await another task to complete or abandon its Obligation.
+    pub async fn transaction_get(
+        self: &Arc<Self>,
+        request_headers: &HeaderMap,
+        ok_to_wait: bool,
+    ) -> (Option<Arc<CacheData>>, Option<Obligation>) {
+        let mut sub = self.0.subscribe();
+
+        loop {
+            let mut awaitable = false;
```
aturon, line 139: It'd be helpful to have a brief comment here giving the semantics / invariants for this variable.

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+                    .filter_map(|v| key_objects.objects.get(v))
+                    .collect();
+                // First, if we raced to produce an obligation, defer in favor of the other.
+                if response_keyed_objects.iter().any(|data| data.obligated) {
+                    return false;
+                }
+
+                // Done dealing with other obligations; now we just deal with results that we have.
+                // Pick a fresh result if we now have it:
+                // fresh or stale.
```
aturon, line 198: I think this comment line is a leftover?

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+            if awaitable && ok_to_wait {
+                let _ = sub.changed().await;
+                continue;
+            }
+
+            // There's nothing fresh in the cache, and no obligation we can wait on.
+            // Take the write lock with the intention of generating our own obligation.
+            let mut obligated: Option<Obligation> = None;
+            let mut data: Option<Arc<CacheData>> = None;
+            self.0.send_if_modified(|key_objects| {
```
aturon, line 178: It's worth calling out here that you're _always_ going to return `false`, i.e. use this as a silent modification.

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+                    v.obligated = false;
+                }
+            }
 
-        let entry = cache_key_objects.objects.entry(variant).or_default();
-        let mut response_object = entry.inner.lock().unwrap();
-        response_object.transactional = TransactionState::Present(object);
-        response_object.generation += 1;
+            if !cache_key_objects.vary_rules.contains(meta.vary_rule()) {
+                // Insert at the front, run through the rules in order, so we tend towards fresher
```
aturon, line 267: Should we also _move_ the rule to the front if it already exists?

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+            self.0.send_if_modified(|key_objects| {
+                // Now under the write lock.
+                // We might have a stale result, or someone else might have an obligaton;
+                // pick the best we can.
+                let response_keys: Vec<_> = key_objects
+                    .vary_rules
+                    .iter()
+                    .map(|v| v.variant(request_headers))
+                    .collect();
+                let response_keyed_objects: Vec<_> = response_keys
```
aturon, line 187: Worth a comment: "these are the existing cache entries that are potentially valid for this request, if any."

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
+                    .into_iter()
+                    .filter_map(|cache_value| cache_value.present.as_ref())
+                    .filter(|cache_data| cache_data.meta.is_fresh())
+                    .next()
+                {
+                    data = Some(Arc::clone(fresh));
+                    return false;
+                }
+
+                // Finally, generate an obligation based on the most recent vary rule
```
aturon, line 209: Have we confirmed this is the behavior in production? I'm thinking of a scenario like:

* Two requests come in for the same URL, but with different headers
* Cache contains:
  * Vary rules: `[["header1"], ["header2"]]`
  * Stale responses keyed as: `"header1: foo"` and `"header2: bar"`
* First request has `header1: foo` and `header2: bar`; we mark the `header1: foo` object as obligated
* Second request has no `header1`, but has `header2: bar`
  * **Thus, we generate a _second_** obligation, even though the one for the second request may well apply.

We could reasonably expect these two requests to be collapses, instead.

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
         }
+    }
 
-        let body = CollectingBody::new(body);
-        let variant = meta.variant();
-        let object = Arc::new(CacheData { body, meta });
+    /// Insert into the corresponding entry.
+    ///
+    /// If a clear_obligation is provided, clear the "obligated" bit on that Variant in the same
+    /// transaction (so there's only one wakeup). Note the clear_obligation variant may differ from
```
aturon, line 248: It'd be helpful to spell out _why_ the variants may differ: because we place an obligation marker based on the _existing_ vary rules, but the actual response being inserted may provide a different rule.

lib/src/cache/store.rs
```diff
@@ -99,98 +99,262 @@ impl From<WriteOptions> for ObjectMeta {
...
-    /// No data, no obligation to fetch
-    #[default]
-    Missing,
-    /// The metadata is present in the cache; the content is available, possibly only as streaming
-    /// content.
-    Present(Arc<CacheData>),
+impl Obligation {
+    /// Fulfill the obligation by providing write options and a body.
+    pub fn complete(mut self, options: WriteOptions, body: Body) {
+        let mut request_headers = HeaderMap::default();
```
aturon, line 327: Alternatively:

```rust
let request_headers = std::mem::take(&mut self.request_headers);
```

lib/src/wiggle_abi/cache.rs
```diff
@@ -223,15 +238,41 @@ impl FastlyCache for Session {
...
+        // Look up once, joining the transaction only if obligated:
+        let e = cache.transaction_lookup(&key, &headers, false).await;
+        let ready = e.found().is_some() || e.go_get().is_some();
+        // If we already got _something_, we can provide an already-complete PeekableTask.
+        // Otherwise we need to spawn it and let it block in the background.
+        let task = if ready {
+            PeekableTask::complete(e)
+        } else {
+            PeekableTask::spawn(Box::pin(async move {
+                Ok(cache.transaction_lookup(&key, &headers, true).await)
```
aturon, line 254: The double lookup here is slightly unfortunate, but ultimately harmless -- and treating this "atomically" would require a bit more contortion, so, 👍

aturon (commented):
Here's another batch of comments / questions.

lib/src/cache/store.rs
cceckman-at-fastly: Done in 54d0a1f8

lib/src/cache/store.rs
cceckman-at-fastly: Done in 2bf2139

lib/src/cache/store.rs
cceckman-at-fastly: Done in 525a1f9

lib/src/cache/store.rs
cceckman-at-fastly: Good thought! I'll add a test for this case, double-check some of the cached code, and circle back.

lib/src/cache/store.rs
cceckman-at-fastly: Done in 525a1f9

lib/src/cache/store.rs
cceckman-at-fastly: Phrasing this slightly differently as I work through it:

If our heuristic for "which variant will apply to the next request" is "what is the last Vary rule received", we would *not* expect for these requests to be collapsed - because they have distinct `header1` values. We'd be erring on the side of "less latency, more requests" by not collapsing them (i.e. not blocking the second request on the fulfillment of the first).

(Per your other comment, my implementation isn't actually using the "most recent received" heuristic, but that's somewhat separate.)

cceckman-at-fastly: Wrote a test in 04d9b0e that matches your example (`header1` and `header2` are swapped, but I'm pretty sure it's the same behaviorally.) I used the `cceckman/outside-test` branch to run it on compute platform.

This implementation matches the compute platform behavior: the requests are not collapsed.

Presumably the cache code authors of ECP had the "most recent Vary is right" heuristic in mind as well? I can see it getting pretty bad if there's an old `Vary: Accept` or something that trumps a `Vary: Authorization` or something, and causes ~all requests to serialize.

aturon: Thanks for checking this out -- that makes good sense to me! 🙏

aturon (approved):
Fantastic work! Very happy to see this land ✨

PR #472, opened 2025-04-23
Cache API: User metadata

cceckman-at-fastly:
Provide insert & lookup handlers for user metadata.

### DIFF
--- a/lib/src/wiggle_abi/cache.rs
+++ b/lib/src/wiggle_abi/cache.rs
@@ -421,7 +421,7 @@ impl FastlyCache for Session {
         let len: u32 = md_bytes
             .len()
             .try_into()
-            .map_err(|_| crate::Error::CacheError(crate::cache::Error::TooLargeUserMetadata))?;
+            .expect("user metadata must be shorter than u32 can indicate");
         if len > user_metadata_out_len {
             memory.write(nwritten_out, len)?;
             return Err(Error::BufferLengthError {
--- a/lib/src/wiggle_abi/cache.rs
+++ b/lib/src/wiggle_abi/cache.rs
@@ -2,6 +2,7 @@ use core::str;
 use std::sync::Arc;
 use std::time::Duration;
 
+use bytes::Bytes;
 use http::HeaderMap;
 
 use crate::body::Body;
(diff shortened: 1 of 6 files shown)


cee-dub (dismissed):
LGTM

lib/src/wiggle_abi/cache.rs
```diff
@@ -401,7 +412,30 @@ impl FastlyCache for Session {
...
+        let entry = self.cache_entry(handle.into()).await?;
+
+        let md_bytes = entry
+            .found()
+            .map(|found| found.meta().user_metadata())
+            .ok_or(crate::Error::CacheError(crate::cache::Error::Missing))?;
+        let len: u32 = md_bytes
+            .len()
+            .try_into()
+            .map_err(|_| crate::Error::CacheError(crate::cache::Error::TooLargeUserMetadata))?;
```
aturon, line 424: We should probably panic here, rather than setting up a new error variant, since this situation should be impossible: we don't support streamed writing of the user metadata, so the known up-front length will always be a u32.

aturon (commented):
LGTM modulo one nit!

lib/src/wiggle_abi/cache.rs
cceckman-at-fastly: Done in 469a0b3.

The component version talks about `list<u8>`, not e.g. `ptr, u32`; would that turn into a u64 in a universe where we support wasm64? (Not that that will happen, just wondering :smile: )

test-fixtures/src/bin/cache.rs
```diff
@@ -354,6 +356,36 @@ fn test_vary_combine() {
...
+    let mut writer = insert(key.clone(), Duration::from_secs(10))
+        .user_metadata(Bytes::copy_from_slice(b"hi there"))
+        .execute()
+        .unwrap();
+
+    // Body not yet written, but we should be able to read the metadata right away.
+    {
+        let fetch = lookup(key.clone()).execute().unwrap();
+        let Some(got) = fetch else {
+            panic!("did not fetch streaming")
```
cee-dub, line 371: Why `panic` vs. `.expect()`?

cee-dub (approved):
LGTM! Just a question for my own curiosity.

cee-dub (approved):

PR #474, opened 2025-05-06
:straight_ruler:  Known length option for core cache

cceckman-at-fastly:
Support `known_length` insert option.

Generate an error on body read if the `known_length` provided does not match the total number of bytes written.
(The fixture test passes on the compute platform and in Viceroy.)

### DIFF
--- a/test-fixtures/src/bin/cache.rs
+++ b/test-fixtures/src/bin/cache.rs
@@ -3,7 +3,7 @@
 use bytes::Bytes;
 use fastly::cache::core::*;
 use fastly::http::{HeaderName, HeaderValue};
-use std::io::Write;
+use std::io::{Read, Write};
 use std::time::Duration;
 use uuid::Uuid;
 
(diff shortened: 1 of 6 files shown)


test-fixtures/src/bin/cache.rs
```diff
@@ -21,12 +21,32 @@ fn main() {
...
 }
 
+/// Wait for the length of a cached object to be known.
+///
+/// Internally in Viceroy, all writes to cached bodies are streaming writes, and are processed
+/// concurrently. That means an `insert..finish()` sequence, followed by a `lookup..to_stream()`,
+/// may still observe the body as "streaming" for a period of time.
+///
+/// If we didn't provide a known .length() along with the insert data, we can sleep+poll until the
+/// length is known, implying that the whole bosy is available.
```
aturon, line 41:
```suggestion
/// length is known, implying that the whole body is available.
```

aturon (approved):
💯, no notes except a typo

test-fixtures/src/bin/cache.rs
cceckman-at-fastly: Done

aturon (approved):

PR #477, opened 2025-05-12
Store `ExecuteCtx` reference in `Session` in place of multiple `Arc`s

ulyssa:
I was going to add another argument to `Session::new` locally, but realized that at this point it would probably be better to just pass in an `Arc<ExecuteCtx>`. This PR replaces the cloned `Arc` references that get passed to `Session::new` with a single clone of `Arc<ExecuteCtx>`, and updates the `Session` methods to access things through it instead.

Since I've updated `ExecuteCtx::new` to now return an `Arc`, I have moved the `.with_*(mut self, ...)` methods out of `ExecuteCtx` in e08aed3 and into an explicit builder type with a `.finish()` method that returns the wrapped `Arc`, so that the `cli/` code can still use them in a straightforward way.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -89,19 +89,19 @@ pub struct ExecuteCtx {
     /// An almost-linked Instance: each import function is linked, just needs a Store
     instance_pre: Arc<Instance>,
     /// The acls for this execution.
-    acls: Arc<Acls>,
+    acls: Acls,
     /// The backends for this execution.
-    backends: Arc<Backends>,
+    backends: Backends,
     /// The device detection mappings for this execution.
-    device_detection: Arc<DeviceDetection>,
+    device_detection: DeviceDetection,
     /// The geolocation mappings for this execution.
-    geolocation: Arc<Geolocation>,
+    geolocation: Geolocation,
     /// Preloaded TLS certificates and configuration
     tls_config: TlsConfig,
     /// The dictionaries for this execution.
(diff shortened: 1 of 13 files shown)


lib/src/execute.rs
```diff
@@ -114,7 +115,7 @@ pub struct ExecuteCtx {
     /// The secret stores for this execution.
     secret_stores: Arc<SecretStores>,
     /// The shielding sites for this execution.
-    shielding_sites: Arc<ShieldingSites>,
+    shielding_sites: ShieldingSites,
     /// The cache for this service.
     cache: Arc<Cache>,
```
cceckman-at-fastly, line 120: I think this can also be a Cache instead of `Arc<Cache>`?

cceckman-at-fastly: Nevermind - it does, so that we can generate additional `Arc`s for some of the independent futures.

cceckman-at-fastly (dismissed):

lib/src/execute.rs
ulyssa: It need to be an `Arc` so that it can be cloned and passed into an `async` block in places like here: https://github.com/fastly/Viceroy/blob/da2ce4eefdb8f24edd86db8ed38d06d1eaf7a1b3/lib/src/wiggle_abi/cache.rs#L270-L272

That code could potentially be updated to pass around a clone of the `Arc<ExecuteCtx>` instead, but I've left it as-is for now.

PR #478, opened 2025-05-15
Improve TLS certificate loading, handling and validation

jedisct1:
Improved Certificate Loading:
  - Now fails explicitly when no certificates can be loaded
  - Properly reports when CA certificates are unavailable

Better Certificate Validation:
  - Certificate validation checks during connection establishment
  - Added basic validation of certificate data to ensure certificates aren't empty
  - Added proper error handling for invalid DNS names

Proper TLS Connection Validation:
  - Added explicit error handling for TLS connection failures
  - Added logging for certificate validation failures for easier debugging


lib/src/upstream.rs
```diff
@@ -153,17 +168,45 @@ impl hyper::service::Service<Uri> for BackendConnector {
...
+                        Error::IoError(std::io::Error::new(
+                            std::io::ErrorKind::Other,
+                            format!("TCP connection error: {}", e),
+                        )),
+                    )))
+                }
+            };
 
-            let remote_addr = tcp.peer_addr()?;
+            let remote_addr = match tcp.peer_addr() {
```
cceckman-at-fastly: I don't think this hunk does anything-

The `?` operator applies one or more `from`/`into` conversions between the error types. `crate::error::Error` implements `From<std::io::Error>` (it's hidden in the little `#[from]` note on the `IoError` declaration), so this is just doing (more verbosely) what `?` does.

cceckman-at-fastly (dismissed):
Patched with some syntactic changes - `map_err` and use of the `?`, mostly - to cut out some of the error-handling boilerplate.

@jedisct1 If this looks good to you, it's good to merge IMO!

src/upstream.rs
```diff
@@ -49,16 +49,24 @@ impl TlsConfig {
...
+                if !cert_added {
+                    return Err(Error::TlsNoCertsAdded);
+                }
             }
             Err(err) => return Err(Error::BadCerts(err)),
         }
         if roots.is_empty() {
-            warn!("no CA certificates available");
+            return Err(Error::TlsNoCAAvailable);
         }
```
kailan, line 61: I know some of this was pre-existing but I believe we can simplify to:

```rust
let certs = rustls_native_certs::load_native_certs().map_err(Error::BadCerts)?;
let mut roots = rustls::RootCertStore::empty();
let (added, failed) = roots.add_parsable_certificates(&certs.into_iter().map(|c| c.0).collect::<Vec<_>>());
if failed > 0 {
    warn!("failed to load {} certificate(s). attempting to continue with {} available certificate(s)", failed, added);
}
if roots.is_empty() {
    return Err(Error::TlsNoCAAvailable);
}
```

src/upstream.rs
```diff
@@ -154,7 +167,9 @@ impl hyper::service::Service<Uri> for BackendConnector {
         };
 
         Box::pin(async move {
-            let tcp = connect_fut.await.map_err(Box::new)?;
+            let tcp = connect_fut
+                .await
+                .map_err(|e| std::io::Error::other(format!("TCP connection error: {}", e)))?;
```
kailan: The HTTP connector already provides descriptive error messages so we probably don't need this extra wrapping


```suggestion
            let tcp = connect_fut.await?
```

src/upstream.rs
kailan: This also means the `TlsNoCertsAdded` error variant can be removed

kailan:
@jedisct1 I hope you don't mind that I took the liberty of reviving this pull request and taking advantage of some since-released Rust features that can simplify the error construction. I'll re-request review from both parties and hopefully we can ship your changes asap.

kailan (dismissed):
Chatted with @jedisct1 out of band – we're happy to go ahead with this considering Charles' earlier review.

kailan (approved):

PR #481, opened 2025-05-27
stale-while-revalidate

cceckman-at-fastly:
Implement `stale-while-revalidate` behavior.

During the `stale-while-revalidate` period (after the "fresh" period), the response will not be fresh (`!is_fresh`) but will be usable (`is_usable`). In this window, the first transaction to get a stale response will also get an obligation to `GoGet` the fresh response, which they can fulfil by `insert`ing a new body or by `update`ing the metadata of the existing item.

While working on this, fix some tests with unintended concurrency: use `finish` instead of `flush` in some cases.


lib/src/cache.rs
```diff
@@ -159,6 +164,24 @@ impl CacheEntry {
...
     }
+
+    /// Insert the provided body into the cache.
+    pub fn insert(&mut self, options: WriteOptions, body: Body) -> Result<(), crate::Error> {
+        let go_get = self.take_go_get().ok_or(Error::NotRevalidatable)?;
+        go_get.insert(options, body);
+        Ok(())
+    }
+
+    /// Freshen the existing body according to the provided options.
```
aturon, line 175: Not a blocker, but the wording here could be a bit misleading. Maybe: `Freshen the existing cache item according to the provided options, without changing the body`

lib/src/cache.rs
```diff
@@ -159,6 +164,24 @@ impl CacheEntry {
...
+    pub fn update(&mut self, options: WriteOptions) -> Result<(), crate::Error> {
+        let go_get = self.take_go_get().ok_or(Error::NotRevalidatable)?;
+        let Err((go_get, err)) = go_get.update(options) else {
+            return Ok(());
+        };
+        // On failure, preserve the obligation.
+        self.go_get = Some(go_get);
+        Err(err)
```
aturon, line 183: Style nit: I think this would be easier to read with a `match`, despite the extra indent level. Not a blocker.

aturon (approved):
Fantastic work, as usual. ✨ 

Only tiny nits -- feel free to ignore, it's fine as-is.

PR #483, opened 2025-05-28
Ensure body reads only return 0-length at the end

### LINKED ISSUE #479: Zero-length chunk in Body generates premature body end (set_auto_decompress_gzip bug)
I'm using [Request set_auto_decompress_gzip](https://docs.rs/fastly/latest/fastly/struct.Request.html#method.set_auto_decompress_gzip) but oddly on my laptop with `fastly compute serve` the decompressed body has no length most of the time. It seems to work fine in Compute.

Here's the code I'm running: https://fiddle.fastly.dev/fiddle/211577df, although in Fiddle it is broken differently, always returning `body_len: 0`.

If I run it locally with `fastly compute serve` I mostly get an incorrect body length of 0 but sometimes get a correct body length of 3410.

When this isn't working, Chunk::CompressedHttpBody only receives 10 bytes, the gzip header, but no deflate stream: [31, 139, 8, 0, 0, 0, 0, 0, 0, 3].

I think I have narrowed this down and will work on a pull request.

cceckman-at-fastly:
Tests and a slightly broader preventative measure for #479.

The Rust [`Read`](https://doc.rust-lang.org/std/io/trait.Read.html) trait uses a return value of `Ok(0)` to indicate a (permanent/temporary) end-of-stream, e.g. end-of-file.
We use the same convention for the `fastly_http_body::read` hostcall.

The `Body` [trait](https://docs.rs/http-body/latest/http_body/trait.Body.html) is a little more careful, with a separate `is_stream_end` method distinct from the "length of frame" property.

However, this means if our `Body` implementations _ever_ yield a zero-length chunk, the `fastly_http_body::read` hostcall will look like an end-of-stream to the guest. That's at the root of #479.

#480 addresses the observed cause- that the gzip decoder in particular can yield zero-length chunks when the header + body are split. However, it's in principle possible for another (current or future) variant of `Body` to reintroduce the issue.

-   Test for zero-length bodies, as an integraton test and as a property test. If #480 is patched out, both of these groups of tests reproduce the issue with the gzip body.
-   Tackle the issue where it occurs, at the conversion from the `http::Body` stream-style API to the `std::io::Read`-style API. Add a `Body::read` method that acts more like the `AsyncRead` traits; use it from the hostcalls, instead of putting `Body` internals in those hostcalls.

Fixed #479 (to my satisfaction)


acme (approved):
👍 Thanks. I'm not very familiar with the codebase, but this looks good.

PR #484, opened 2025-05-30
Surrogate keys in core cache API

cceckman-at-fastly:
Support for surrogate keys and purging thereby.

Check that we consume all the options for `insert`-like operations
and `purge`, so if an SDK starts using an option Viceroy doesn't understand, it fails visibly.

- **Add tests for surrogate keys**
- **Implement surrogate keys & non-soft purge, remove TODOs**
- **Self-check that all insert options are handled**
- **Implement soft purge**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,12 +74,11 @@ package-check:
 	rm -rf .cargo/
 	rm -rf verify-publishable/
 
-# Re-generate the adapter, and move it into `lib/adapter`
+# Regenerate the adapter, and move it into `lib/data`.
 .PHONY: adapter
 adapter:
 	cargo build --release \
 		-p viceroy-component-adapter \
 		--target wasm32-unknown-unknown
-	mkdir -p lib/adapter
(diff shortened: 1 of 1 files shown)


test-fixtures/src/bin/cache.rs
```diff
@@ -646,6 +648,123 @@ fn test_stale_while_revalidate() {
...
+        key
+    }
+
+    let key1 = write_key(["keyA", "keyB"], "value1");
+    let key2 = write_key(["keyA"], "value2");
+    let key3 = write_key(["keyB"], "value3");
+
+    fastly::http::purge::soft_purge_surrogate_key("keyB").unwrap();
+    // Allow either of two behaviors:
+    // - Does not return stale results
```
aturon, line 694: Why do we want to allow the first behavior, given the TTL?

aturon (approved):
Fixes look great, :shipit:!

PR #489, opened 2025-06-03
Update Viceroy to the latest WIT files.

dgohman-fastly:
This updates Viceroy to the WIT files and component adapter as of Executed 4319d1e, and updates the host implementations to match the WIT interfaces.

This also makes the adapter a separate workspace, and updates the build rules accordingly, to more closely match how this work in the ExecuteD repository.


dgohman-fastly:
cc @erikrose

lib/wit/deps/fastly/compute.wit
```diff
@@ -422,14 +515,17 @@ interface http-req {
...
     streaming: bool,
   ) -> result<pending-request-handle, error>;
 
-  send-async-streaming: func(h: request-handle, b: body-handle, backend: string)
--> result<pending-request-handle, error>;
+  send-async-streaming: func(
+    h: request-handle,
+    b: body-handle,
+    backend: string
+  ) -> result<pending-request-handle, error>;
```
cee-dub, line 528: Any chance to make these parameter and/or function names more consistent? The change from `h` & `b` in one, while adding `h` & `b` to the other is unfortunate.

Also `send-async-streaming` is parallel to `send-async-v2` while the names may lead one to believe that `send-async-streaming` is modeled after `send-async`.

lib/wit/deps/fastly/compute.wit
cee-dub, line 1: Love to see the comments!

crates/adapter/src/fastly/core.rs
```diff
@@ -3905,3 +3967,709 @@ pub mod fastly_shielding {
...
+
+    #[export_name = "fastly#init"]
+    pub fn init(abi_version: u64) -> FastlyStatus {
+        fastly_abi::init(abi_version)
+    }
+}
+
+/// The env module is all alternative names for existing functions. This is the equivalent to the
+/// linker aliases in xqd-codegen.
+mod env {
```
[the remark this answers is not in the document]
dgohman-fastly: It's a historical artifact at this point.

It goes all the way back to Emscripten, which used "env" as the name of a module which represented the C global namespace. I imagine people didn't find it confusing at the time because Web pages don't have environment variables. As Wasm evolved, the "env" module name stuck around as the default module name in a lot of Wasm tools, because it ends up getting baked into a lot of places so it's annoying to change.

When Fastly first started experimenting with Wasm, there were a lot of big-picture questions to answer, so bikesheding the default "env" name that the tools came with was not a priority. When it came time to make a customer-facing product, they did change "env" to prettier names like "fastly_http_body" and "fastly_cache" and so on, however by that time, "env" was already baked into compiled Wasm code that we had to continue supporting.

So "env" in the adapter here is just for compatibility with existing Wasm programs that were compiled in the early days of what we now call Fastly Compute. New programs compiled today don't use it. And it's not present at all in the WIT interfaces that we're transitioning to.

lib/wit/deps/fastly/compute.wit
dgohman-fastly: At this point, ExecuteD is the upstream source for these, so I'll submit a separate PR updating these names there first, to make it easier to keep everything in sync.

I myself am not super familiar with these APIs, so I'd appreciate any other suggestions you have for naming changes to help make the API understandable!

lib/wit/deps/fastly/compute.wit
cee-dub: Great to see them come across from the wit(x).

crates/adapter/src/fastly/core.rs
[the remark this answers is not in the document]
cee-dub: Great! That sounds like an improvement.

dgohman-fastly:
Closing this for now; I'm working on a new PR which will subsume these.

PR #493, opened 2025-07-01
Support execute_and_stream_back

cceckman-at-fastly:
I had missed this; but some of the "range request" tests uncovered that I had missed it.

- Address several lints
- Sync tests with internal copy
- Test and implement insert_and_stream_back


aturon (approved):
LGTM -- lots of pleasant cleanups in this one ✨

PR #497, opened 2025-07-19
feat: Add Pushpin proxying support for local testing

harmony7:
This PR adds Viceroy code and hostcall implementations to support the hostcalls `fastly_http_req#redirect_to_grip_proxy` and `fastly_http_req#redirect_to_grip_proxy_v2`.

It adds a new runtime command-line flag, `--local-pushpin-port=<port>` that enables this feature.

## What does this PR do?

This pull request introduces the foundational support within Viceroy for proxying requests to a local Pushpin instance. This enables developers to locally test Fastly Compute services that use real-time features like HTTP streaming or WebSockets-over-HTTP, which are powered by Fanout/Pushpin.

Specifically, this change implements the host-side logic for the following ABI functions: `fastly_http_req#redirect_to_grip_proxy` and `fastly_http_req#redirect_to_grip_proxy_v2`

When a guest module calls one of these functions, Viceroy will now intercept the request and proxy it to a configured Pushpin server instead of handling it directly.

The configuration of Pushpin is performed by a corresponding update to the Fastly CLI, which will set up the appropriate routes and start Pushpin during the `fastly compute serve` command.

## Why is this change important?

Currently, there is no way to locally test Compute features that rely on Fanout. This is a significant gap in the local development workflow, forcing developers to deploy their services to test real-time functionality. By adding this proxying capability to Viceroy, we bridge that gap and create a much more complete and efficient local development experience.

## How is this implemented?

A new `pushpin.rs` Module: A new module is introduced to encapsulate all the proxying logic.

If the guest code calls `redirect_to_grip_proxy` (or `_v2`), the request is replayed to the Pushpin instance, using the name of the backend as the route ID (sets the `Pushpin-route` header value to the backend name). For example, a request to `/api/user` to a backend named `origin` will be sent to Pushpin as `/api/user` with the header `Pushpin-route: origin`. Fastly CLI configures Pushpin with the `accept_pushpin_route=true` setting.

During this forwarding, the original body of the request must be replayed from the beginning. To enable this, I've added a new `body_tee.rs` module, designed to allow forking a stream.

For WebSocket upgrades, the underlying connection is then handed off to a `copy_bidirectional` task to stream data between the client and Pushpin. In order to allow upgrading the original request, code has been added at the beginning of `handle_request` to save an `on_upgrade` future.

## Details

* A new "error" enumeration, `NonHttpResponse`, has been added, which enables the receiver to signal to `handle_request` that the guest wants to return a non-http response, allowing the handler to perform handling specific to the non-http response type.
* A new enumeration `DownstreamResponse` has been added that represents the response itself, which may be an `Http(Response<Body>)` or a `RedirectToPushpin(PushpinRedirectInfo)`. The hostcall for `redirect_to_grip_proxy` (and `_v2`) sends this `RedirectToPushpin` response type, allowing the receiver to signal `NonHttpResponse` to the handler.
    * The existing `DownstreamResponse` enumeration has been renamed to `DownstreamResponseState` to better reflect its name and to mirror other uses within Fastly.
* `handle_request` adds code to handle `NonHttpResponse` response types. If the wrapped response is `RedirectToPushpin`, then if this feature is enabled, `proxy_through_pushpin` is invoked, proxying the request to Pushpin and streaming the result, and upgrading to WebSocket if the response is `101 Switching Protocols`.
* If a backend definition has a path segment, then it is respected by the Pushpin `replace_beg` configuration, but that setup is performed in the CLI, not in Viceroy.
* This has a [companion PR in the Fastly CLI](https://github.com/fastly/cli/pull/1509) which will invoke Pushpin with the appropriate routes before starting Viceroy.

### DIFF
--- a/lib/src/execute.rs
+++ b/lib/src/execute.rs
@@ -11,7 +11,7 @@ use {
             Backends, DeviceDetection, Dictionaries, ExperimentalModule, Geolocation,
             UnknownImportBehavior,
         },
-        downstream::{prepare_request, DownstreamMetadata, DownstreamRequest},
+        downstream::{prepare_request, DownstreamMetadata, DownstreamRequest, DownstreamResponse},
         error::ExecutionError,
         linking::{create_store, link_host_functions, ComponentCtx, WasmCtx},
         object_store::ObjectStores,
@@ -360,6 +360,17 @@ impl ExecuteCtx {
         &self.tls_config
     }
 
+    async fn maybe_receive_response(
+        receiver: oneshot::Receiver<DownstreamResponse>,
+    ) -> Option<(Response<Body>, Option<anyhow::Error>)> {
+        if let Ok(resp) = receiver.await {
+            return match resp {
+                DownstreamResponse::Http(resp) => Some((resp, None)),
+            };
+        }
+        None
+    }
+
     /// Asynchronously handle a request.
     ///
     /// This method fully instantiates the wasm module housed within the `ExecuteCtx`,
@@ -430,7 +441,7 @@ impl ExecuteCtx {
             let original = std::mem::replace(&mut downstream.sender, sender);
             let (resp, err) = self.spawn_guest(downstream, receiver).await;
             let resp = guest_result_to_response(resp, err);
-            let _ = original.send(resp);
+            let _ = original.send(DownstreamResponse::Http(resp));
         });
     }
 
@@ -466,7 +477,11 @@ impl ExecuteCtx {
                 Ok(()) => {
                     // Drop lock and wait for the guest to process our request.
                     drop(reusable);
-                    return (receiver.await.unwrap_or_default(), None);
+
+                    if let Some(response) = Self::maybe_receive_response(receiver).await {
+                        return response;
+                    }
+                    return (Response::default(), None);
                 }
                 Err(nr) => next_req = nr,
             }
@@ -483,7 +498,7 @@ impl ExecuteCtx {
     async fn spawn_guest(
         self: Arc<Self>,
         downstream: DownstreamRequest,
-        receiver: oneshot::Receiver<Response<Body>>,
+        receiver: oneshot::Receiver<DownstreamResponse>,
     ) -> (Response<Body>, Option<anyhow::Error>) {
         let active_cpu_time_us = Arc::new(AtomicU64::new(0));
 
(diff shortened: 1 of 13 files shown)


lib/src/execute.rs
```diff
@@ -406,14 +435,54 @@ impl ExecuteCtx {
...
+        info!("response status: {:?}", resp.status());
+
+        if let Some(e) = err {
+            match e.downcast::<NonHttpResponse>() {
+                Ok(NonHttpResponse::PushpinRedirect(redirect_info)) => {
+                    let backend_name = redirect_info.backend_name;
+                    let redirect_request_info = redirect_info.request_info;
+                    info!("Pushpin redirect signaled to backend '{}'", backend_name);
+
+                    let local_pushpin_proxy_port = match local_pushpin_proxy_port {
```
ulyssa, line 452: I don't think you need the clone at the start of the method. Since it should be `Copy`, I believe you can just make this:

```suggestion
                    let local_pushpin_proxy_port = match self.local_pushpin_proxy_port {
```

ulyssa (changes requested):
This overall looks good to me, but I'm not sure whether the body teeing behaviour actually matches what h2o is doing in practice, although teeing might be close enough.

harmony7:
Thanks @ulyssa, I've applied your suggestions. Please review again =)

harmony7:
> I'm not sure whether the body teeing behaviour actually matches what h2o is doing in practice, although teeing might be close enough.

For this I went with what would have to happen because the original request's body needs to be replayed in full to the proxied request.

harmony7:
Because of recent changes in the main branch, I did a rebase and merged the conflicts.

ulyssa:
> > I'm not sure whether the body teeing behaviour actually matches what h2o is doing in practice, although teeing might be close enough.
>
> For this I went with what would have to happen because the original request's body needs to be replayed in full to the proxied request.

I talked with Fred about this some last week, and it seems like there were some plans for sending bodies to `pushpin` in Compute at one point that never fully materialized. From what I can tell, it's not doable today, but I'm going to keep looking into the current status. Doing the tee'ing in Viceroy for now seems fine, and we can adjust it to be closer to what the originally planned behaviour for Compute and Fanout was in the future once that's actually implemented.

ulyssa (approved):

harmony7:
> I talked with Fred about this some last week, and it seems like there were some plans for sending bodies to `pushpin` in Compute at one point that never fully materialized. From what I can tell, it's not doable today, but I'm going to keep looking into the current status. Doing the tee'ing in Viceroy for now seems fine, and we can adjust it to be closer to what the originally planned behaviour for Compute and Fanout was in the future once that's actually implemented.

Thanks! that sounds great =) I appreciate your thoroughness

PR #502, opened 2025-08-06
Use shielding definitions from Fastly SDK

cceckman-at-fastly:
Strike the TODO; we're already depending on an SDK release with
shielding support.


acw (approved):
:tada:

PR #506, opened 2025-08-08
Release 0.14.2

cceckman-at-fastly:
- Update version number in Cargo.toml and lockfiles
- Update changelog
- Upgrade to wasmtime v28 and Rust 1.85, to fix Windows double panic; and update dependencies to bring lockfiles up to date for publication


ulyssa (approved):

cceckman-at-fastly:
This still didn't make it to crates.io. Publishing with `cargo publish` or `cargo publish --locked` forced a regeneration of the lockfile (yes, even when locked).

Seems like this is a bug- https://github.com/rust-lang/cargo/issues/13986 maybe?

ulyssa (approved):

PR #510, opened 2025-08-11
Upgrade to 1.84, switch to resolver 3, and upgrade packages

cceckman-at-fastly:
- **Upgrade Rust to 1.84, switch to version-aware resolver**
- **Regenerate lockfiles at 1.84**


cceckman-at-fastly:
See #512 for solution

Issue #507, opened 2025-08-11
Panic on windows using Rust 1.84


cceckman-at-fastly:
We're a little stuck at the moment:

- `cargo publish` requires an up-to-date lockfile. (`--locked` means "the lockfile is already supposed to be up to date", not "don't change it".)
- Updating the lockfile upgrades `spdx`, a transitive dependency
- The new `spdx` version 0.10.9 requires Rust >= 1.85 for `edition2024`
- Running tests on Windows with Rust 1.85 errors: [see here](https://github.com/fastly/Viceroy/actions/runs/16839945146)

So... we need to get past Rust 1.85 somehow before we can cut a new release.

cceckman-at-fastly:
From [here](https://github.com/fastly/Viceroy/actions/runs/16890122181/job/47847532032?pr=508), the `trap-test`:

```
running 2 tests

thread 'tokio-runtime-worker' panicked at library\core\src\panicking.rs:218:5:
panic in a function that cannot unwind
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library\std\src\panicking.rs:692
   1: core::panicking::panic_nounwind_fmt
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library\core\src\intrinsics\mod.rs:3869
   2: core::panicking::panic_nounwind
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library\core\src\panicking.rs:218
   3: core::panicking::panic_cannot_unwind
             at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library\core\src\panicking.rs:307
   4: enum2$<wasmtime::runtime::func::HostContext>::array_call_trampoline<viceroy_lib::linking::WasmCtx,wasmtime::runtime::func::impl$56::into_func::closure_env$0<viceroy_lib::linking::WasmCtx,viceroy_lib::wiggle_abi::fastly_http_resp::add_to_linker::closure_env
             at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\wasmtime-25.0.3\src\runtime\func.rs:2225
   5: _CxxFrameHandler3
   6: is_exception_typeof
   7: _C_specific_handler
   8: is_exception_typeof
   9: _CxxFrameHandler3
  10: _chkstk
  11: RtlUnwindEx
[...]

PR #511, opened 2025-08-11
Try rust 1.89

cceckman-at-fastly:
Roll all the way up to `stable`, see if that also reproduces the Windows crash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,6 +39,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
+version = "0.2.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


ulyssa:
It's not quite the same (we're seeing `exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN`, instead of `0xc0000028`/`STATUS_BAD_STACK`), but I'm wondering if this is similar to bytecodealliance/wasmtime#10289, and if we can fix the issue by upgrading to `windows-2025` like in bytecodealliance/wasmtime#10290.

cceckman-at-fastly:
Maybe - I do think it's from some of the unwinding stuff that came in 1.84 (https://github.com/fastly/Viceroy/issues/507).

I had wondered how upstream got around it, but hadn't found https://github.com/bytecodealliance/wasmtime/pull/10290, nice!

PR #513, opened 2025-08-12
Update to Wasmtime 35.

dgohman-fastly:
Update to Wasmtime 35, WASI 0.2.6,
wasmparser/wasm-encoder/wit-component 0.236.0, cranelift-entity 0.122.0, wit-bindgen-rust-macro 0.43.0, and fastly-shared to 0.11.5.

This removes the `UnknownImportBehavior::ZeroOrNull` option for now, because the way it works in Wasmtime changed, and perhaps we don't need it anymore.


rust-toolchain
cceckman-at-fastly, line 1: If updating the `rust-toolchain`, we should also bring the `.github/workflow` files to the same version, yes? Making sure we're testing what we say

cceckman-at-fastly (approved):

dgohman-fastly:
@cee-dub Zero-or-null behavior is something we'd like to move away from anyway, as it doesn't make sense for components. Would it work for your use case to use `UnknownImportBehavior::Trap` instead?

@cceckman-at-fastly Rebased to fix merge conflicts.

dgohman-fastly:
Great, thanks!

PR #518, opened 2025-09-02
Reorganize directory structure

dgohman-fastly:
This moves the main code out of `lib/src` and into a top-level `src`, moves the adapter to its own workspace, and adds a wasm_abi directory and moves the wit, witx, and adapter directories into it. There should be no functional change here; this just makes the directory structure more convenient to work with.


ulyssa (approved):
Big +1 from me! :shipit: 

I always found the `lib/src/` structure a little odd when switching to doing things in this repo. Mirroring the XQD directory structure more closely will also be nice when comparing the repos.

PR #529, opened 2025-09-20
Add support for building the library adapter.

dgohman-fastly:
The library adapter is similar to the main component adapter, but doesn't include the exports. This is a building block for component libraries.


chenyan2002:
I wonder if we need the `--library` flag. Can Viceroy figure out if the customer needs the library adapter or the full adapter by looking at the exports?

PR #532, opened 2025-09-30
Fix double borrows in `get_with_by_ref()` calls

dmartin:
Removes one layer of borrows from keys passed to Moka's `get_with_by_ref()`, restoring compatibility with the trait bound changes in Moka 0.12.11.


deg4uss3r:
Great fix!

PR #534, opened 2025-10-01
Update to the latest WITs and adapter.

dgohman-fastly:
Changes:
 - `on-behalf-of` and `service-id` strings moved out of `fastly:compute` and into `fastly:adapter`.
 - Trailers now use a dedicated `trailer-error` type that can indicate when trailers aren't available yet.
 - `pending-request` is now `pending-response` and `request-promise` is now `pending-request`, for consistency with the rest of the API.
 - `send-error-detail` is now a variant with richer fields.
 - Some options records gained some new options.
 - `secret-store.plaintext` no longer returns an `option`, because it wasn't needed.


chenyan2002 (approved):
LGTM

PR #538, opened 2025-10-06
Update adapter with the memory shift fix

chenyan2002:
Propagating changes from https://github.com/fastly/ExecuteD/pull/5528.

The adapter module takes a fixed two pages of memory. It gets this memory either via cabi_realloc exported from the main module or memory.grow at the beginning of the component instantiation. The is not a safe assumption for managed languages like Go or Python.

This PR avoids this problem by rewriting the main module to always access the memory with a 2 page offset. In this way, the first two pages of the memory is never touched by the main module. Then the adapter can safely write its state and stack to the first two pages without corrupting user memory.


dgohman-fastly (approved):
Thanks!

PR #540, opened 2025-10-07
Return 5XX for requests that land in a reused session that crashes

ulyssa:
This PR moves the logic for turning an error into a 500 response to happen right after the guest crashes by calling `Session::close_downstream_response_sender` so that it will still apply to reused sessions. Fortunately, the ways that a guest can exit here are much simpler than in Compute, so this ends up being a simpler change.


src/execute.rs
```diff
@@ -986,17 +996,24 @@ fn write_profile(store: &mut wasmtime::Store<WasmCtx>, guest_profile_path: Optio
...
-            .unwrap()
+    err.as_ref().map(anyhow_response).unwrap_or(resp)
+}
+
+fn exec_err_to_response(err: &ExecutionError) -> Response<Body> {
+    if let ExecutionError::WasmTrap(e) = err {
+        anyhow_response(e)
     } else {
-        resp
+        panic!("failed to run guest: {err}")
```
acw, line 1006: Why panic here, rather than turning the execution error into an anyhow response?

acw (approved):
Seems fine, although I have one question about a panic.

src/execute.rs
ulyssa: I'm doing this to match the behaviour from here:

https://github.com/fastly/Viceroy/blob/f2ea186b044082dcd806e2dfa1450fc4beef7f01/src/execute.rs#L592-L606

PR #545, opened 2025-10-18
Allow components to have multiple memories.

dgohman-fastly:
Viceroy currently uses arbitrary values for the number of instances and tables allowed in components. Use a similarly arbitrary number of memories as well.


erikrose (approved):
Looks good. No config necessary unless somebody's really going to town! :-D

PR #549, opened 2025-10-27
fix: use original static backend host

posborne:
Prior to this change, using a static backend without explicitly setting a Host header or configuring an override resulted in the expect case being hit.  The configured non-override static backend URI host should be used.

---

I encountered this while working directly against the WIT definitions but this implementation is shared with code paths on the wiggle abi side of things and in a cursory look it seems like it might experience similar issues, though it could be that I missed a key piece in either the SDKs or other ABI code.

This is the code snippet from the experimental python WIT bindings that would hit the failure.  The static backend was configured in viceroy (but with no additional override_host set).

```python
request = http_req.Request.new()
request.set_method("GET")
request.set_uri(path)

# TODO: without this, we hit a panic in viceroy in send
request.insert_header("Host", b"localhost")

body = http_body.new()
response, response_body = http_req.send(request, body, backend_name)
```

### DIFF
--- a/src/upstream.rs
+++ b/src/upstream.rs
@@ -228,7 +228,13 @@ fn canonical_host_header(
                 .authority()
                 .and_then(|auth| HeaderValue::from_str(auth.as_str()).ok())
         })
-        .expect("Could determine a Host header")
+        .or_else(|| {
+            backend
+                .uri
+                .host()
+                .and_then(|h| HeaderValue::from_str(h).ok())
+        })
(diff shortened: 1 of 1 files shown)


posborne:
I'm going to merge this for now as it matches the current xqd behavior and avoids the host panic; I think we need to make a final determination on whether we should have `set_uri` reject URIs without a host/authority.  This has not been the case at the ABI layer historically but has been at the Rust SDK layer, for instance.

PR #551, opened 2025-11-10
Add support for manual HTTP framing headers

kailan:
Currently, services that make use of the `framing_headers_mode_set` hostcalls for requests and responses fail to run on Viceroy, throwing an error.

This pull requests implements these hostcalls by attaching metadata to the relevant request or response object, which is then picked up in the `upstream` or `downstream` module when the request or response is sent.

### DIFF
--- a/src/component/compute/http_req.rs
+++ b/src/component/compute/http_req.rs
@@ -1,11 +1,7 @@
 use {
     crate::{
         component::{
-            bindings::fastly::compute::{
-                http_body, http_req, http_resp,
-                http_types::{self, FramingHeadersMode},
-                types,
-            },
+            bindings::fastly::compute::{http_body, http_req, http_resp, http_types, types},
             compute::headers::{get_names, get_values},
         },
         error::Error,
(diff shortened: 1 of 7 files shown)


src/component/compute/http_req.rs
```diff
@@ -438,15 +438,29 @@ impl http_req::HostRequest for ComponentCtx {
...
+        h: Resource<http_req::Request>,
         mode: http_types::FramingHeadersMode,
     ) -> Result<(), types::Error> {
-        match mode {
-            http_types::FramingHeadersMode::ManuallyFromHeaders => {
-                Err(Error::NotAvailable("Manual framing headers").into())
+        let manual_framing_headers = match mode {
+            FramingHeadersMode::ManuallyFromHeaders => true,
+            FramingHeadersMode::Automatic => false,
+        };
```
franfastly: I'd rather be consistent with `FramingHeadersMode`'s path. 

I'm leaning more towards always using `http_types::FramingHeadersMode`, since we do the same with other items in `http_types`, but both options are Ok to me.

franfastly (commented):
This is a minor comment

PR #552, opened 2025-11-10
Core cache: lock FOUND to USABLE

cceckman-at-fastly:
The Compute platform, at the moment, only returns FOUND if within the USABLE period (in HTTP semantics, "fresh" or "stale-while-revalidate"). Unfortunately, SDKs rely on this property, and check FOUND or USABLE variously; sometimes (e.g. in Rust), this results in a `FOUND`-but-not-`USABLE` result being returned.

We can fix SDKs going forward, but not existing usage. Match Compute's behavior by locking `FOUND` to `USABLE`. If we want to return expired results, we'll need more / other bits.

Also add tests that exercise this behavior. (Really they're testing the Rust SDK's handling of this behavior, but, better than not having it.)

### DIFF
--- a/src/component/compute/cache.rs
+++ b/src/component/compute/cache.rs
@@ -503,13 +503,16 @@ impl api::HostEntry for ComponentCtx {
 
         let mut state = api::LookupState::empty();
         if let Some(found) = entry.found() {
-            state |= api::LookupState::FOUND;
-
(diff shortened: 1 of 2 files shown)


amsantavicca (approved):
LGTM! 🥇

PR #589, opened 2026-03-01
Reapply changes to support rich integration tests for fastly compute

dpetrick:
This PR extends Viceroy's public API to support richer integration testing by introducing several new capabilities or extending existing ones:
- In-memory backend handlers (`InMemoryBackendHandler` trait + `Handler` wrapper): Backends can now be configured with custom async handlers that intercept requests and return responses without making real HTTP calls, enabling fully in-process integration tests.
- Similarly, there's now a dynamic backend registration interceptor (yes, incredible name... `DynamicBackendRegistrationInterceptor` trait): Allows test harnesses to hook into dynamic backend registration and modify or augment backends as they are registered at runtime (e.g., to attach in-memory handlers).
- Add Display trait to `Cache` and related structs to allow pretty printing. This is useful for debugging potentially incorrect cache state that the application produces.
- Extend `Cache` with `purge_all` convenience method.
- `GuestHandle` + `run_to_completion()`: The `handle_request` and `spawn_guest` call chains now return a `GuestHandle` so callers can await full guest completion (including async logging, cache writes, etc.) after receiving the HTTP response.
- Logging endpoint monitoring (`EndpointsMonitor` / `EndpointListener`): Provides a channel-based mechanism to register listeners on named logging endpoints, allowing tests to capture and assert on log output from the guest.
- `ExecuteCtx::fork()`: Allows creating a lightweight fork of an execution context that shares the compiled WASM engine/module but has fresh per-test state (cache, request IDs, pending reuse), making it efficient to run many tests against the same compiled guest.
- Re-exported dependencies: async_trait, http, and hyper are now re-exported from the crate root so downstream consumers don't need to depend on matching versions.

### DIFF
--- a/cli/tests/integration/common.rs
+++ b/cli/tests/integration/common.rs
@@ -418,8 +418,8 @@ impl Test {
                     .await
                     .map(|result| {
                         match result {
-                            (resp, None) => resp,
-                            (_, Some(err)) => {
+                            (resp, None, _) => resp,
+                            (_, Some(err), _) => {
                                 // Splat the string representation of the runtime error into a synthetic
                                 // 500. This is a bit of a hack, but good enough to check for expected error
                                 // strings.
--- a/cli/tests/integration/common/backends.rs
+++ b/cli/tests/integration/common/backends.rs
@@ -76,6 +76,7 @@ impl TestBackends {
                 grpc: false,
                 client_cert: None,
                 ca_certs: vec![],
+                handler: None,
             };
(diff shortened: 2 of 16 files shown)


kpfleming:
Apologies for this PR going stale... there's a lot of great stuff here, and much of it aligns with some changes I had been planning on making for similar reasons. I'll break this PR up into smaller pieces which focus on each new feature and move those forward.

PR #593, opened 2026-03-23
Support guest profiling for components

posborne:
This change follows with the work I did in upstream wasmtime to support guest profiling for components:
https://github.com/bytecodealliance/wasmtime/pull/10507

Integration tests for both components and core modules were added to ensure we have some coverage of the guest profiling (just ensuring it generates profiler output).

### DIFF
--- a/cli/tests/integration/common.rs
+++ b/cli/tests/integration/common.rs
@@ -92,6 +92,8 @@ pub struct Test {
     via_hyper: bool,
     unknown_import_behavior: UnknownImportBehavior,
     adapt_component: bool,
+    profiling_strategy: ProfilingStrategy,
+    guest_profile_config: Option<viceroy_lib::GuestProfileConfig>,
 }
 
 impl Test {
@@ -116,6 +118,8 @@ impl Test {
             via_hyper: false,
             unknown_import_behavior: Default::default(),
             adapt_component: false,
+            profiling_strategy: ProfilingStrategy::None,
+            guest_profile_config: None,
         }
     }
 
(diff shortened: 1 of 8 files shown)


cli/tests/integration/common.rs
```diff
@@ -287,6 +293,12 @@ impl Test {
         self
     }
 
+    /// Enable guest profiling with the specified configuration.
+    pub fn with_guest_profiling(mut self, config: viceroy_lib::GuestProfileConfig) -> Self {
+        self.guest_profile_config = Some(config);
+        self
+    }
```
dgohman-fastly, line 300: Should there also be a `with_profiling_strategy` method for setting the `profiling_strategy` field?

posborne: Yeah, I had plumbed that through but didn't write additional tests for those bits (as my changes are isolated to guest profiling).  The naming and plumbing of some of these config bits is a bit confusing as the profiling strategy is unique to native profiling (not guest).

I have some changes now that consolidate things down to a single `profiling_config` which consolidates the guest/native/none paths in a single field.  It's a fairly large change in terms of files touched, so I'm inclined to land this change as-is and then separately open up that change for review separately as it is beyond the scope of what we're changing here.

dgohman-fastly (approved):
Makes sense; thanks!

cli/tests/integration/common.rs
posborne: https://github.com/fastly/Viceroy/pull/598

PR #598, opened 2026-04-01
Refactor profiling configuration to use unified ProfilingConfig enum

posborne:
Replace the confusing dual profiling configuration system (ProfilingStrategy + GuestProfileConfig) with a single, unified ProfilingConfig enum that cleanly distinguishes between native profiling (PerfMap/JitDump/VTune) and guest profiling (Firefox Profiler format).


dgohman-fastly (dismissed):
Sorry for the delay here; this looks good to me, though it'll need a rebase now.

dgohman-fastly (approved):

PR #600, opened 2026-04-14
Adapter shim for pollable incompatibility

cceckman-at-fastly:
There are many handle types that, in WITX, are valid for `async_io::select`; among them are cache handles, cache replace handles, and HTTP cache handles. Each of these is backed with an async item (well, not HTTP cache handles, because they are not yet implemented in Viceroy) and can be waited on by `select`. The module ABI doesn't have strong notions of typing; these u32s are all in the same space.

In WIT, everything that can be waited on by `async-io.select` is an alias for the `async-io.pollable` handle type. The "component" version of Wasmtime checks these handles by type; not all u32s are the same.
But `cache-entry`, `cache-replace-entry`, and `http-cache-entry` are not such aliases; they are their own resources. If you pass one of these handles to `async-io.select`, you'll get an error.

We realized this leads to an incompatibility in adapted code. Code that works as a module (passing e.g. `cache-entry` to `select`) will break under the adapter, prior to this change.

This adds a shim layer to the adapter, tagging the three handle types above with extra bits in the high nybble. (These bits are reserved for purposes like this.) The adapter sets these bits per-type when producing the handle, and removes them before calling down in to the component ABI.

When these incompatible handles arrive at `select` (or `is_ready`), the adapter strips the type bits out to recover the handle, then uses it to call `subscribe`. `subscribe` returns the same handle, but tracked as a `pollable`, so it can then be fed in to the underlying ABI call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Add adapter shim for `async-io.select`-incompatible handles. ([#600](https://github.com/fastly/Viceroy/pull/600))
 - Add stub implementations for resvpnproxy hostcalls.
 
 ## 0.16.5 (2026-03-23)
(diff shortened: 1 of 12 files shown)


dgohman-fastly:
This ended up evolving into changes that are now included in https://github.com/fastly/Viceroy/pull/637.

PR #601, opened 2026-04-16
Add options for experimenting with wasm gc and exceptions.

dgohman-fastly:
Add an argument to `viceroy_lib::ExecuteCtx::build` for enabling additional Wasm language features, and add command-line arguments to the viceroy CLI for enabling gc and exceptions.


kailan (commented):
This looks ok, but does enabling `CM_GC` enable the component model? It's unclear from the wasmtime docs.

dgohman-fastly:
> This looks ok, but does enabling `CM_GC` enable the component model? It's unclear from the wasmtime docs.

Component model support is [already enabled by default](https://docs.rs/wasmtime/39.0.0/wasmtime/struct.Config.html#method.wasm_component_model), so `CM_GC` is just adding the GC support in addition to that.

kailan (dismissed):
> Component model support is [already enabled by default](https://docs.rs/wasmtime/39.0.0/wasmtime/struct.Config.html#method.wasm_component_model), so `CM_GC` is just adding the GC support in addition to that.

Fantastic, thank you!

kailan (approved):

PR #603, opened 2026-04-20
Address clippy warnings

kailan:
For as long as I can remember, `cargo clippy` has produced a huge amount of warnings in this project. This pull request resolves as many of those warnings as possible, with a few exceptions:

- Large size difference between enum variants
- Functions with too many arguments
- Complex type definition

I felt these were out of scope for a large sweeping change like this so I've annotated them to be ignored by clippy.

As a result, clippy output is now clean, and it can be added to the CI workflow to ensure new changes are checked for best practices.

### DIFF
--- a/cli/tests/integration/body.rs
+++ b/cli/tests/integration/body.rs
@@ -52,7 +52,7 @@ viceroy_test!(zero_length_raw_chunks_are_transparent, |is_component| {
                             };
                             tokio::task::yield_now().await;
                         }
-                        let _ = write.send_trailers(HeaderMap::default());
+                        let _ = write.send_trailers(HeaderMap::default()).await;
                     });
                     Response::builder()
                         .status(StatusCode::OK)
@@ -103,7 +103,7 @@ viceroy_test!(gzip_breaks_are_ok, |is_component| {
                             };
                             tokio::task::yield_now().await;
                         }
-                        let _ = write.send_trailers(HeaderMap::default());
+                        let _ = write.send_trailers(HeaderMap::default()).await;
                     });
                     Response::builder()
                         .status(StatusCode::OK)
--- a/cli/tests/integration/common.rs
+++ b/cli/tests/integration/common.rs
@@ -347,7 +347,7 @@ impl Test {
 
         let ctx = ExecuteCtx::build(
             &self.module_path,
-            self.profiling_strategy.clone(),
+            self.profiling_strategy,
             HashSet::new(),
             self.guest_profile_config.clone(),
             self.unknown_import_behavior,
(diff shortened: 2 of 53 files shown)


cli/tests/integration/common.rs
```diff
@@ -522,6 +522,7 @@ impl Drop for TestServer {
 }
 
 #[derive(Clone)]
+#[allow(clippy::type_complexity)]
```
cceckman-at-fastly: Is there an expected follow-up simplifying this? I hate to leave it "just" disabled without a plan for follow-up

src/wiggle_abi/req_impl.rs
```diff
@@ -808,7 +808,7 @@ impl FastlyHttpReq for Session {
     ) -> Result<(), Error> {
         let req = self.request_parts_mut(req_handle)?;
 
-        let version = hyper::Version::try_from(version)?;
+        let version = hyper::Version::from(version);
```
kpfleming, line 811: I'm not sure about this change.

cli/tests/integration/body.rs
```diff
@@ -52,7 +52,7 @@ viceroy_test!(zero_length_raw_chunks_are_transparent, |is_component| {
                             };
                             tokio::task::yield_now().await;
                         }
-                        let _ = write.send_trailers(HeaderMap::default());
+                        let _ = write.send_trailers(HeaderMap::default()).await;
```
kpfleming, line 55: This looks a bit odd; the return value is discarded, but was previously not being `await`ed so the Future was never being polled.

cli/tests/integration/body.rs
```diff
@@ -103,7 +103,7 @@ viceroy_test!(gzip_breaks_are_ok, |is_component| {
...
```
kpfleming, line 106: Same concern as line 55.

kpfleming (dismissed):
Overall there are a ton of improvements here, including use of if-let instead of single-arm match, use of let chains instead of nested if blocks, and similar things. I've left a couple of comments on changes that I'm not sure I fully understand and which will need review from others.

There are probably more improvements like this to be had if we allow Viceroy to be built with a newer version of Rust, it's still pinned to 1.90.

kailan:
> There are probably more improvements like this to be had if we allow Viceroy to be built with a newer version of Rust, it's still pinned to 1.90.

I switched to Rust 1.95 and ran clippy there, which found a few more improvements that I included in this PR. For making that upgrade, I have #604 open which is based on this branch.

src/wiggle_abi/req_impl.rs
cceckman-at-fastly: There's an existing explicit `From` for these types, so the `try_from` is a blanket impl with `Error=Infallible`.

This is fine 👍

cceckman-at-fastly:
Generally looks good.

I'd rather get the ignored lints fixed than allowed; https://github.com/fastly/Viceroy/pull/607 addresses those.

cli/tests/integration/body.rs
kailan: I looked into this and it doesn't affect the outcome of the test, so I think it's ok to keep it this way.

cli/tests/integration/body.rs
kailan: See response above

kpfleming (approved):

PR #604, opened 2026-04-20
Upgrade to Rust 1.95

kailan:


kpfleming:
Would you consider making a slightly different set of changes? With `rust-toolchain.toml` in place, there shouldn't be any need to specify a Rust version anywhere else (in configuration files or workflows), and there should not be any need to invoke `rustup` to install, uninstall, or set `default` either. That change would result in the version being specified in only one location and eliminate the risk of version mismatches.

We don't need to set `rust-version` in `cli/Cargo.toml` since that crate is not intended for re-use; for the top-level `Cargo.toml` (the `viceroy-lib` crate) the `rust-version` sets the MSRV for the crate, it does not specify the compiler version to be used when building it. That means it should be set to some reasonably recent version which is compatible with the source code, but doesn't need to be any newer.

If you'd prefer, I'd be happy to make the changes in this branch instead.

kailan:
> Would you consider making a slightly different set of changes? With `rust-toolchain.toml` in place, there shouldn't be any need to specify a Rust version anywhere else

I'll give it a go!

Dockerfile
```diff
@@ -18,9 +18,7 @@ RUN apt-get update \
 ENV LD_LIBRARY_PATH=/usr/local/lib
 
 # Install Rust, rustfmt, and the wasm32-wasi cross-compilation target
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.52.1 -y
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain stable -y
```
kpfleming, line 21: No harm making this change, but it doesn't appear that this Dockerfile is being used anywhere in the repository and it just builds a very simple image containing rustup and Rust.

kpfleming (approved):

PR #607, opened 2026-04-23
Address Clippy lint on variant size by boxing in NextRequest

cceckman-at-fastly:
Instead of `allow`ing the variance in #603, fix it.


kailan (approved):
Thank you!

Issue #536, opened 2025-10-02
Stop committing `viceroy-component-adapter.wasm`


erikrose:
We currently commit a built WASI p1-to-p2 adapter in `wasm_abi/data/`. I suspect it's just because our build system isn't rigged to automatically build it before building the rest of Viceroy (which needs to read it in at compile time using `include_bytes!()`).

Let's contrive to make our build system build the adapter automatically, perhaps with a `build.rs`. This will avoid…

1. Me forgetting to update the adapter before building Viceroy and then wondering why my changes didn't take effect
2. Having a nondeterministic and thus unverifiable binary committed. (It turns out the binaries differ _a lot_ based on whether they're built on Linux or the Mac (imports and functions being in different orders, making diffing next to impossible); IHNI about Windows. Even between 2 Linux boxes, they differ in their symbol names suffixes, e.g. $_ZN25viceroy_component_adapter6macros10eprint_u3215eprint_u32_impl17hc6b497a227a95847E.llvm.**6107973161718614711**.)

PR #614, opened 2026-04-25
Automate compilation of component adapter

kailan:
Previously, the compiled binaries were checked in to the repo. This pull request adds a build script to compile them locally when the source is updated, resolving #536


kailan:
Reverting to draft since this is causing issues with `cargo package`

WhiteAbeLincoln:
Replaced by #720

PR #618, opened 2026-05-05
Fix noshift adapter selection for wit-bindgen library components

### LINKED ISSUE #619: Bug: noshift adapter not selected for wit-bindgen library (export-only) components
Follow-up to #582 (and the closed ExecuteD#6130 it implemented).

`viceroy adapt` picks the noshift adapter when a module has wit-bindgen *imports* (the case #582 covered). A library component that only *exports* a wit-bindgen interface still receives the shifted adapter. When such a library is composed with a noshift consumer via `wac plug`, the canonical-ABI memcpy crosses between shifted and unshifted memory views and silently corrupts pointer-backed values.

Symptoms observed against a minimal `wac plug`-composed inspector + consumer pair (consumer imports a `request-info` record, inspector exports the function returning a variant):

| WIT type | Behaviour |
|---|---|
| `u32` | round-trips |
| bare variants (no payload) | round-trip |
| `string` | length correct, contents zeroed/stale |
| `list<u8>` (in) | length correct, contents zeroed |
| `list<u8>` (out) | length 0 |

ExecuteD#6130 framed a *complete* fix as complicated. This case is structurally symmetric to the import-side fix in #582 and amenable to the same scoped-partial-fix approach (heuristic on the export side, by canonical-ABI mangled name).

Fixed in #618.

ebonura-fastly:
## Summary

Extends the noshift adapter selection in `viceroy adapt` to detect wit-bindgen *exports*, not only imports. Library components that only export a wit-bindgen interface previously received the shifted adapter, which mismatches the consumer's noshift adapter at the canonical-ABI boundary and silently corrupts pointer-backed values across `wac plug` composition.

Closes #619.

## Background

[#582](https://github.com/fastly/Viceroy/pull/582) by @chenyan2002 fixed the import-side wit-bindgen case from [fastly/ExecuteD#6130](https://github.com/fastly/ExecuteD/issues/6130). The export-only library case wasn't covered, because the motivating example for #582 (kv-store-component used through xqd-codegen, not through `wac plug`) didn't surface it. Once you compose two adapted components via `wac plug`, the symmetric case shows up: the consumer module imports a wit-bindgen interface and gets noshift; the library module only exports the interface and still gets shifted; canonical-ABI memcpy lands at the wrong physical address by 2 pages.

#6130 framed a *complete* fix (shifting addresses through canonical-ABI lift/lower for nested data structures) as complicated. This change is not that complete fix. It applies the same scoped-partial-fix pattern from #582 symmetrically to the export side.

## Reproducer

A barebones `wac plug`-composed inspector + consumer pair (consumer imports a `request-info` record and gets back an `action` variant; inspector exports the function). Symptoms before this patch:

| WIT type | Behaviour |
|---|---|
| `u32` | round-trips |
| bare variants (no payload) | round-trip |
| `string` | length correct, contents zeroed/stale |
| `list<u8>` (in) | length correct, contents zeroed |
| `list<u8>` (out) | length 0 |

After this patch, all of the above round-trip correctly, both locally under `viceroy serve` and through the production path on a real Compute service.

## Implementation

```rust
let needs_no_shift_adapter = has_wit_bindgen_imports(bytes) || has_wit_bindgen_exports(bytes);
```

`has_wit_bindgen_exports` walks the export section and looks for canonical-ABI mangled names (containing both `:` and `#`, e.g. `fastly:component-template/inspector@0.1.0#inspect`). Standard wasip1 exports (`_start`, `__main_void`, `memory`, `cabi_realloc`, `cabi_realloc_wit_bindgen_*`) contain neither character, so the heuristic has no false positives in practice.

Pure WASIp1 modules continue to use the shifted adapter so its perf benefit is preserved.

## Validation

- `make ci` clean (116 tests + 2 doc tests pass; clippy and fmt clean).
- `cargo test -p viceroy-lib adapt::` — 5/5 pass (3 heuristic detection tests + 2 end-to-end `adapt_bytes` wiring tests asserting producer-metadata flag strings).
- Built and ran the companion component template (in `fastly/edge-storage`, branch pending push access) end-to-end with `cargo build --target=wasm32-wasip1` + `viceroy adapt` + `wac plug`. Adapted library reports `viceroy adapt [0.17.1 (library, noshift)]`, adapted consumer reports `viceroy adapt [0.17.1 (noshift)]`.
- Deployed the resulting `plugged.wasm` to a real Fastly Compute service ([cc-template-smoketest-6178.edgecompute.app](https://cc-template-smoketest-6178.edgecompute.app)). `GET /hello` returns `200 echoed: GET /hello`; `GET /admin` returns `403 admin paths are blocked`. The production path through `copa` + ExecuteD doesn't invoke the buggy adapter (per chenyan2002's analysis in #6130), so this change has no impact on Cypress behaviour.

## Discussion

I'd value review on whether this is the right shape for the fix. It's possible there's nuance specific to the export side I haven't spotted; happy to adjust if so.

## Related

- Tracking issue: #619
- Slack thread in #compute-components: https://fastly.slack.com/archives/C06S9HTEES2/p1777993828111399
- Project tracking: [MCSOC-3504](https://fastly.atlassian.net/browse/MCSOC-3504)
- Companion PR: pending push access on `fastly/edge-storage` for the component template that surfaced this bug

[MCSOC-3504]: https://fastly.atlassian.net/browse/MCSOC-3504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/adapt.rs
```diff
@@ -28,10 +28,33 @@ pub fn adapt_wat(wat: &str) -> anyhow::Result<Vec<u8>> {
 
 /// Given bytes that represent a core wasm module, adapt it to a component using the viceroy
 /// adapter.
+///
+/// The adapter has two flavors: the default *shifted* one, which moves the user
+/// module's memory up by 2 pages so the adapter can park scratch state in the
+/// first 2 pages without needing `cabi_realloc`; and the *noshift* one, which
+/// allocates state via the user module's `cabi_realloc` and leaves memory
+/// untouched. The shifted form is faster (state lives at a fixed offset, no
```
chenyan2002: The shifted form is not faster. It's more universal, but it doesn't work with wit-bindgen. Anyway, I feel like it's too much detail in the comments, which may confuse the reader, instead of helping the user to understand the code.

src/adapt.rs
```diff
@@ -188,3 +216,171 @@ fn has_wit_bindgen_imports(bytes: &[u8]) -> bool {
...
+/// mangled form `<package>:<interface>@<version>#<func>`, i.e. a wit-bindgen
+/// export consumed via cross-component composition. Such a module cannot use
+/// the shifted adapter because its `cabi_realloc` would return shifted-view
+/// pointers that wasmtime would then write source data to in the wrong
+/// physical location.
+///
+/// Detection is by structural pattern: any export name containing both `:` and
+/// `#`. The standard wasip1 exports a Rust Compute service emits (`_start`,
+/// `__main_void`, `memory`, `cabi_realloc`, `cabi_realloc_wit_bindgen_*`)
+/// contain neither, so the check has no false positives in practice.
```
chenyan2002:
```suggestion
// This is a best effort implementation to check if the export name has a WIT ABI mangled name of the form `<package>:<interface>#<func>`. The user module may mimic the naming convention without using wit-bindgen, but that's not supported by the Fastly SDK and the user doesn't get any benefit by minicing the export name.
```

chenyan2002 (commented):
Mostly LG. 

My main concern is that the comments seem to be generated by an AI, and it's not very accurate. It would suggest removing all the generated comments. 

Likewise, the unit tests are also somewhat artificial, and doesn't reflect the real use case. If you want, you could include your library binary to make sure the code doesn't regress on your use case. I'm also fine to not add any new tests. This is mainly a bug fix that we forget to check the export side of the module when deciding which adapter to use.

ebonura-fastly:
@chenyan2002 Thanks, Yan. I cleaned up the comments and force-pushed an update. About the longer block, I had used AI-generated notes to help orient myself in this part of the adapter code, but you’re right they were too much detail for the final PR so I toned them down massively.

I also removed the artificial WAT-based tests from the pushed branch for now.

For the test coverage question, I tried your real-fixture suggestion locally. I rebuilt the actual inspector library Wasm from the component-template work and confirmed it reproduces the export-side shape this PR is fixing. The fixture would be about 37 KB if we strip nonessential metadata before committing it. With that fixture, the test verifies that Viceroy chooses the `library, noshift` adapter path for a real wit-bindgen-generated library component.

Would you prefer that I reintroduce test coverage using that binary fixture, or keep this PR smaller with no new tests?

src/adapt.rs
```diff
@@ -188,3 +192,26 @@ fn has_wit_bindgen_imports(bytes: &[u8]) -> bool {
 
     false
 }
+
+// Best-effort check for WIT ABI-mangled export names, such as
```
chenyan2002:
```suggestion
// Best-effort check for wit-bindgen export names, such as
```

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## Unreleased
 
-- Rename 'session' to 'sandbox' since that is the preferred term in public documentation. ([#617](https://github.com/fastly/Viceroy/pull/617))
+- Fix noshift adapter selection for wit-bindgen library components. The existing heuristic only checked for wit-bindgen *imports*; an export-only library component still received the shifted adapter, silently corrupting strings and lists when composed with another adapted component via `wac plug`. The heuristic now also detects canonical-ABI mangled exports. ([#618](https://github.com/fastly/Viceroy/pull/618)) See also [#582](https://github.com/fastly/Viceroy/pull/582) and [ExecuteD#6130](https://github.com/fastly/ExecuteD/issues/6130).
```
chenyan2002:
```suggestion
- Fix noshift adapter selection for wit-bindgen library components. ([#618](https://github.com/fastly/Viceroy/pull/618))
```

This is enough for user-facing changelog

chenyan2002 (commented):
Looks good!

As for the tests, if you already have the binary test ready, and it's not too much code, feel free to add it. Otherwise, we can keep the PR as it is.

ebonura-fastly:
@chenyan2002 I added the real binary fixture test.

The Wasm fixture is generated from the component-template work I wrote in `fastly/edge-storage`, specifically `components/template/inspector-component`. That component template work is what led me to find this Viceroy bug in the first place: plugging the inspector component into the consumer exposed that an export-only wit-bindgen library component needs the `noshift` adapter path.

Does that provenance sound okay for this PR? Viceroy now has the stripped `.wasm` fixture and the test, but not the source needed to regenerate that exact binary. That source belongs in the `edge-storage` component-template PR. I still only have read access to `fastly/edge-storage`, so I cannot push that branch/PR there directly right now.

Would you prefer that I keep the fixture here as-is, add a short provenance note for where it came from, or wait until the `edge-storage` component-template source is available in a PR?

src/adapt.rs
```diff
@@ -188,3 +192,48 @@ fn has_wit_bindgen_imports(bytes: &[u8]) -> bool {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn adapt_bytes_marks_wit_bindgen_library_fixture_as_library_noshift() {
+        let adapted = adapt_bytes(include_bytes!(
+            "../tests/fixtures/template_inspector_component.wasm"
```
chenyan2002, line 229: Add a short notes to indicate where the wasm comes from

src/adapt.rs
```diff
@@ -188,3 +192,48 @@ fn has_wit_bindgen_imports(bytes: &[u8]) -> bool {
...
+        ))
+        .unwrap();
+
```
chenyan2002, line 232:
```suggestion
// Checking which adapter is used from the custom section.
```
Ideally, we want to read the custom section using `wasmparser`, but this is not too bad.

chenyan2002 (commented):
LGTM! Just some nits.

chenyan2002 (dismissed):
Thanks!

kailan (approved):
Approving based on @chenyan2002's review - many thanks to both of you

PR #621, opened 2026-05-15
Add support for WebSockets passthrough

harmony7:
This PR adds support for WebSockets passthrough ("websocket handoff") in local testing.

It does so by refactoring the existing Pushpin-specific redirection logic into a generalized "handoff" mechanism. This enables Viceroy to support not only Fanout/Pushpin redirects but also direct handoffs to configured backends, specifically enabling previously unimplemented WebSocket proxying capabilities.

This PR is structured into four parts, committed as separate changesets:

## Part 1: Refactor Existing Pushpin Handoff

**Goal:** Abstract the low-level proxy and upgrade logic away from Pushpin-specific structures.

* **New `handoff` Module:** Renamed `src/pushpin.rs` to `src/handoff.rs` and generalized the data structures (`PushpinRedirectInfo` → `HandoffInfo`).

* **Separation of Concerns:** Split the handoff process `perform_handoff()` (formerly `proxy_through_pushpin()`) into smaller phases:
   1. **Request Construction:** `create_request_for_handoff()` handles header merging (respecting protected headers) and initial URI setup.
   2. **URI/Host decoration** after the request is created, adds a Host header and backend in a method appropriate for the handoff mechanism (e.g., the `pushpin-route` header for Pushpin handoff).
   3. **Execution:** `execute_handoff()` connects to the desired handoff target host (e.g., Pushpin), manages the asynchronous connection, and the `101 Switching Protocols` bidirectional I/O upgrade.

## Part 2: Add Handoff to Backend Functionality

**Goal:** Enable the execution engine to route handoffs directly to origin backends.

* **Just another handoff type** Uses the same `perform_handoff()` mechanism as Pushpin, but specifying the backend uri and override host directly.

* **URI & Host Handling:** - Implemented **URI Prepending**: If a backend is configured with a path (e.g., `http://localhost:8080/api`), that path is correctly prepended to the incoming request URI. (Only needed for Backend targets, as this is accomplished using the routes config in Pushpin)
* Implemented **Host Overrides**: The `Host` header is now correctly set based on the backend's `override_host` setting or the backend URI's authority.

## Part 3: Add TLS support for Backend handoff

**Goal:** Enable backend handoff added in part 2 to negotiate with a backend using TLS.

* **Standard TLS Configuration** - Uses the standard TLS configuration from Viceroy as well as from backend configuration to perform connections if a backend's URL schemas include `https`.
* **Optional at the `perform_handoff` level** - so Pushpin integration, designed to work on localhost without TLS, will continue to work correctly

## Part 4: Wire up WebSocket Handoff Hostcalls

**Goal:** Expose the new backend handoff functionality to the Wasm guest via standard Fastly Compute hostcalls.

* **Hostcall Implementation:** Updated `redirect_to_websocket_proxy` and its V2 variant in both the `wiggle_abi` and `component` layers.
* **Feature Completion:** Replaced several `Error::NotAvailable` stubs with actual implementations. Wasm guests can now successfully use `handoff_websocket` (and related calls) to delegate long-lived connections to backends.
* **State Management:** Updated `DownstreamResponseState` to track `HandingOffToBackend`, ensuring proper state transitions and preventing multiple response attempts.

### Technical Notes

* **Protocol Upgrades:** The handoff mechanism explicitly looks for `101 Switching Protocols` and spawns a background task to manage `copy_bidirectional` between the downstream client and the upstream target.
* **Memory Safety:** Moved to owned `String` types for handoff configuration to ensure thread safety and longevity within spawned async tasks, avoiding "escaped function body" lifetime issues.

### How to test it?

Build Viceroy
```
cargo build --release
```

Create an empty Compute project.

In fastly.toml, set up a backend pointing at the public WebSocket.org echo demo server:
```
[local_server]
[local_server.backends]
[local_server.backends.origin]
override_host = "echo.websocket.org"
url = "https://echo.websocket.org:443/"
```

Use something like the following source code
```
/// <reference types="@fastly/js-compute" />
import { createWebsocketHandoff } from "fastly:websocket";

addEventListener("fetch", (event) => event.respondWith(handleRequest(event)));

async function handleRequest(event) {
  const { request } = event;

  if (
    request.method === 'GET' &&
    request.headers.get('upgrade')?.split(',').some(x => x.trim() === 'websocket')
  ) {
    return createWebsocketHandoff(request, "origin");
  }

  return fetch(request, { backend: "origin" });
}
```

Run the app
```
fastly compute serve --viceroy-path=/path/to/Viceroy/target/release/viceroy
```

Connect to the echo server through your app (get wscat if you don't already have it)
```
wscat -e http://localhost:7676/
```


src/execute.rs
```diff
@@ -482,36 +488,116 @@ impl ExecuteCtx {
...
+                    info!("Backend handoff signaled to backend '{backend_name}'");
+
+                    let backend = backends.get(backend_name.as_str());
+                    let backend = match backend {
+                        None => {
+                            error!("Backend handoff signaled to unknown backend '{backend_name}'.");
+                            let resp = Response::builder()
+                                .status(StatusCode::INTERNAL_SERVER_ERROR)
+                                .body(Body::from(hyper::Body::from(
+                                    "Backend handoff signaled to unknown backend '{backend_name}'.",
```
remore: While I was testing this PR, I saw the response body from this line resulted as follows. Looks like string expansion is not working as expected;

> Backend handoff signaled to unknown backend '{backend_name}'

Full curl command;
```
 curl -ksS \
        --header 'connection: upgrade' \
        --header 'upgrade: websocket' \
        --verbose \
        "http://127.0.0.1:7676/ws" ;                       
*   Trying 127.0.0.1:7676...
* Connected to 127.0.0.1 (127.0.0.1) port 7676
> GET /ws HTTP/1.1
> Host: 127.0.0.1:7676
> User-Agent: curl/8.7.1
> Accept: */*
> connection: upgrade
> upgrade: websocket
> 
* Request completely sent off
< HTTP/1.1 500 Internal Server Error
< content-length: 61
< date: Mon, 18 May 2026 06:25:17 GMT
< 
* Connection #0 to host 127.0.0.1 left intact
Backend handoff signaled to unknown backend '{backend_name}'.
```

harmony7: thanks for the catch, it was missing the `format!()` macro

remore: Just confirmed this is fixed now.

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+    host_header: String,
+    display_name: String,
+    path_prefix: Option<String>,
+    extra_headers: Vec<(String, String)>,
+    request_info: Option<HandoffRequestInfo>,
+    orig_request_info: HandoffRequestInfo,
+    orig_body: Body,
+    on_upgrade: OnUpgrade,
+    tls_config: Option<HandoffTlsConfig>,
+) -> Response<Body> {
```
kailan, line 162: This is causing CI failures because the clippy configuration is strict. Perhaps you could move some of these related params into a struct:
```rust
pub struct HandoffConfig {
    pub target_addr: String,
    pub host_header: String,
    pub display_name: String,
    pub path_prefix: Option<String>,
    pub extra_headers: Vec<(String, String)>,
    pub tls_config: Option<HandoffTlsConfig>,
}
```

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        // Resolve SNI Host
+        let cert_host = config
+            .cert_host
+            .as_deref()
+            .unwrap_or(&config.dns_name_fallback);
+        let dnsname = rustls::client::ServerName::try_from(cert_host).unwrap();
+
+        let connector = tokio_rustls::TlsConnector::from(std::sync::Arc::new(client_config));
+        let tls = connector.connect(dnsname, handoff_stream).await.unwrap();
```
kailan: It may be worth at least using `expect` instead of `unwrap` so you can provide a more helpful error message. e.g. `Invalid hostname provided for fanout handoff`

src/execute.rs
```diff
@@ -482,36 +488,122 @@ impl ExecuteCtx {
...
+                        None
+                    };
+
+                    let backend_host = backend_uri
+                        .authority()
+                        .map(|a| match (a.port(), backend_uri.scheme_str()) {
+                            (None, Some("wss")) => format!("{}:443", a),
+                            (None, Some("https")) => format!("{}:443", a),
+                            (None, Some("ws")) => format!("{}:80", a),
+                            (None, Some("http")) => format!("{}:80", a),
```
kailan, line 575: Is it possible to configure backends with ws/wss?

kailan (changes requested):
This looks great – I really like the use of "handoff" instead of "redirect", much more accurate to what is actually happening.

src/handoff.rs
kpfleming: Keep in mind the Rust guidance (https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style) that `expect` messages should describe what you expected to be true.

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+    "proxy-authenticate",
+    "proxy-authorization",
+    "te",
+    "trailer",
+    "transfer-encoding",
+    // Other
+    "cdn-loop",
+];
+
+/// A signal to redirect the request to Pushpin
```
kpfleming: This is a minor point, but the various comments referring to redirection could be updated to handoff instead.

kpfleming: Also references to Pushpin in this module should be made generic unless they are specific to Pushpin.

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+///
+/// - The method and request URI (path + query) are taken from request_info
+///   if provided, or orig_request_info otherwise.
+/// - The body is taken from orig_body.
+/// - Headers are taken from request_info if provided (except for certain protected
+///   headers, which will keep their values from orig_request_info), or
+///   orig_request_info otherwise.
+/// - The Host header is always replaced by the parameter `host_header`.
+/// - `extra_headers`, if provided, are applied. This always replaces existing
+///   headers of the same name.
```
kpfleming: This replacement should not permit overriding protected headers, though, correct? If so, the comment could make that clear.

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+}
+
+/// Creates a request suitable for use with execute_handoff().
+fn create_request_for_handoff(
+    backend_host: &str,
+    redirect_request_info: Option<HandoffRequestInfo>,
+    original_request_info: HandoffRequestInfo,
+    body: Body,
+) -> HttpResult<Request<Body>> {
+    let (path_and_query, method) = if let Some(ref info) = redirect_request_info {
```
kpfleming: Another place where `s/redirect/handoff/` would be applicable.

src/handoff.rs
```diff
@@ -0,0 +1,437 @@
...
+        }
+    }
+}
+
+/// A helper function to build a simple error response.
+fn build_error_response(status: StatusCode, message: impl ToString) -> Response<Body> {
+    Response::builder()
+        .status(status)
+        .body(Body::from(format!("Error: {}", message.to_string())))
+        .expect("Could not build error response")
```
kpfleming: The `expect` message should indicate what was supposed to happen, not what did happen.

kpfleming (changes requested):

src/execute.rs
harmony7: In Viceroy, fastly.toml's local_server lets you put any URL string, and it can include a protocol. If a port is not specified we need a port, so we're setting it based on the protocol. My thought is that when using this feature to forward to a WebSocket backend, it's reasonable for users to expect to be able to configure backend URLs in fastly.toml using the ws/wss schema.

(In contrast, Fastly control panel requires a hostname, a port, and a radio button to select whether to enable TLS.)

src/handoff.rs
harmony7: Actually it _does_ replace protected headers, so I've clarified that.

It's important for user code not to be able to replace protected headers, but `extra_headers` is only set by Viceroy, so it's fine to be able to override them.

src/handoff.rs
harmony7: I've switched to the following, which will never panic:
```
fn build_error_response(status: StatusCode, message: impl ToString) -> Response<Body> {
    let mut resp = Response::new(Body::from(format!("Error: {}", message.to_string())));
    *resp.status_mut() = status;
    resp
}
```

Because this is meant to be called as part of error handling, I find this to reduce mental load of reading the code.

src/execute.rs
kailan: This makes sense to me 👍

kailan (approved):
Looks great, thank you!

kpfleming (approved):

PR #625, opened 2026-05-27
add cargo-binstall metadata for faster installation

syphar:
When using `cargo install` on CI it can take quite some time to install the `viceroy` cli (10/15 minutes somtimes). 

Since we're already publishing the binary in gh releases, we can also use that. 

edit:  https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,18 @@ To install Viceroy as a standalone tool, you'll need to first
 Then run `cargo install --locked viceroy`, which will download and build the latest
 Viceroy release.
 
+### As a prebuilt binary
+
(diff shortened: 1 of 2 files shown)


kailan (approved):
TIL that's how you do that. Thanks very much!

syphar:
forgot to mention the related docs: 
https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md

syphar:
( often `cargo binstall` can also auto-discover this, but sometimes it needs help)

PR #627, opened 2026-05-27
fix: extend transaction_get lookup to observe all pending cache obligations

### LINKED ISSUE #626: Cache: test collapse from later Vary
In talking with @tfreiberg-fastly , we think we came up with an edge case in request collapsing that Viceroy may not implement correctly.

1. Cache contains two variants: `Vary: Header-A`+`Header-A: foo` and `Vary: Header-A`+`Header-A: bar`. Both are expired.
2. Dispatch three transasctions: 1 has `Header-A: foo`, 2 and 3 have `Header-A: bar`. 1 receives a `GoGet`, as does 2; 3 waits on 2 completing.[^simplicity]
3. Transaction 1 completes with an empty vary rule. This means in principle it _can_ fulfill request 3; in practice, request 3 might not get waked yet.
4. Transaction 2 is abandoned.

At this stage, what happens to transaction 3? The desired behavior is that transaction 3 gets a `Found`: there is now a variant that can fulfill this request.

I thought that Viceroy might do the wrong thing, and issue a `GoGet` to transaction 3 instead. However, looking at the code, it looks like that isn't the case: the core loop [here](https://github.com/fastly/Viceroy/blob/75b5675a26698c579736170560e96f0c0f778296/src/cache/store.rs#L203) waits for _any variant_ under same key to be updated, then reviews the whole list.

That said, we should [add a test for this case](https://github.com/fastly/Viceroy/blob/main/test-fixtures/src/bin/cache.rs) for both Compute and Viceroy.

[^simplicity]: if 2 and 3 dispatch concurrently, either might get the `GoGet`; we assume 2 does for simplicity.

THad04:
Closes #626 

## Description
This PR addresses issue #626. The goal was to implement an integration test verifying request-collapsing safety when handling dynamic HTTP `Vary` header evaluations. The test passed, showing that this edge case is covered.

## Changes
1. **Deterministic Test: Added `test_collapse_from_later_vary` to `test-fixtures/src/bin/cache.rs`. It handles explicit host-side handle cleanup via `.cancel_insert_or_update()`.

## Verification
- Verified that all 160 integration tests and 112 unit tests pass 100% green

### DIFF
--- a/test-fixtures/src/bin/cache.rs
+++ b/test-fixtures/src/bin/cache.rs
@@ -1225,36 +1225,30 @@ fn test_collapse_from_later_vary() {
         .unwrap();
     b.finish().unwrap();
 
-    let pending_txn1 = Transaction::lookup(key.clone())
+    let txn1 = Transaction::lookup(key.clone())
         .header(&header_a, "foo")
-        .execute_async()
+        .execute()
         .unwrap();
-    let pending_txn2 = Transaction::lookup(key.clone())
+
+    let txn2 = Transaction::lookup(key.clone())
         .header(&header_a, "bar")
-        .execute_async()
+        .execute()
         .unwrap();
-    assert!(!pending_txn1.pending().unwrap(), "txn1 should have a GoGet");
-    assert!(!pending_txn2.pending().unwrap(), "txn2 should have a GoGet");
 
     let pending_txn3 = Transaction::lookup(key.clone())
         .header(&header_a, "bar")
         .execute_async()
         .unwrap();
     assert!(pending_txn3.pending().unwrap(), "txn3 should be waiting on txn2");
-
-    let txn1 = pending_txn1.wait().unwrap();
-    let txn2 = pending_txn2.wait().unwrap();
     assert!(txn1.must_insert_or_update());
     assert!(txn2.must_insert_or_update());
 
-    // Txn1 completes with an empty vary rule (no .vary_by() on the insert).
-    // This pushes VaryRule::default() to the front of the key's vary_rules and
-    // inserts under the universal variant {}, which matches any request.
     let mut writer = txn1.insert(Duration::from_secs(120)).execute().unwrap();
     writer.write_all(b"the-response").unwrap();
     writer.finish().unwrap();
 
-    // Txn2 is abandoned after Txn1 completes (matches the issue's ordering).
+    // Abandon Txn2 to verify that Txn3 falls back to 
+    // Txn1's wildcard entry instead of spawning a duplicate origin fetch.
     txn2.cancel_insert_or_update().unwrap();
 
     // Txn3 must resolve as Found from Txn1's empty-vary insert.
--- a/test-fixtures/src/bin/cache.rs
+++ b/test-fixtures/src/bin/cache.rs
@@ -1204,9 +1204,13 @@ fn test_collapse_across_vary() {
 }
 
 fn test_collapse_from_later_vary() {
+    // Per issue #626: cache contains two stale variants under Vary: Header-A.
+    // Txn1 (foo), Txn2 (bar), Txn3 (bar) dispatch; Txn1 and Txn2 receive GoGet,
+    // Txn3 waits on Txn2. Txn1 completes with an *empty* vary rule, which can in
+    // principle fulfill Txn3. Txn2 is then abandoned. Txn3 should get Found.
     let key = new_key();
     let header_a = HeaderName::from_static("header-a");
-    // Prefill two stale variants, both under Vary: header-a.
+
     let b = insert(key.clone(), Duration::ZERO)
         .header(&header_a, "foo")
         .vary_by([&header_a])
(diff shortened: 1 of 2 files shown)


test-fixtures/src/bin/cache.rs
```diff
@@ -1202,6 +1203,78 @@ fn test_collapse_across_vary() {
...
+        .unwrap();
+    let pending_txn2 = Transaction::lookup(key.clone())
+        .header(&header_a, "bar")
+        .execute_async()
+        .unwrap();
+    assert!(!pending_txn1.pending().unwrap(), "txn1 should have a GoGet");
+    assert!(!pending_txn2.pending().unwrap(), "txn2 should have a GoGet");
+
+    // Transaction 3 (bar) dispatches and collapses behind Transaction 2.
+    // Sleep briefly so the background task reaches sub.changed() before we proceed.
```
cceckman-at-fastly: Hm - does this fail / flake without the sleep? Or is this proactive/cautionary?

There's a subtlety in Compute that may not be replicated in Viceroy -- that `execute_async` only returns after one round-trip to the cache has occurred. If we do that in Viceroy as well (we may not?) then we should reliably get `.pending` at this point.

test-fixtures/src/bin/cache.rs
```diff
@@ -1202,6 +1203,78 @@ fn test_collapse_across_vary() {
...
+    assert!(pending_txn3.pending().unwrap(), "txn3 should be waiting on txn2");
+
+    let txn1 = pending_txn1.wait().unwrap();
+    let txn2 = pending_txn2.wait().unwrap();
+    assert!(txn1.must_insert_or_update());
+    assert!(txn2.must_insert_or_update());
+
+    // Transaction 2 is abandoned before Transaction 1 completes.
+    // Must use cancel_insert_or_update() — drop() only releases the Rust wrapper;
+    // it does not close the host-side handle or release the Obligation.
```
cceckman-at-fastly: Where's the drop handler for `Transaction`, and what does its stack look like? Are we sure it does not call close?

(Calling `cancel_insert_or_update` here to make it explicit is fine; but I want to double-check this claim.)

cceckman-at-fastly (commented):
Per chat -- I'm actually confused as to how the test passes at HEAD / why the fix fixes it; it looks to me like the fix would collapse when we _shouldn't_? But I suspect I'm missing something

test-fixtures/src/bin/cache.rs
```diff
@@ -1202,6 +1203,72 @@ fn test_collapse_across_vary() {
...
+    b.finish().unwrap();
+
+    let b = insert(key.clone(), Duration::ZERO)
+        .header(&header_a, "bar")
+        .vary_by([&header_a])
+        .execute()
+        .unwrap();
+    b.finish().unwrap();
+
+    let pending_txn1 = Transaction::lookup(key.clone())
```
cceckman-at-fastly: For txn1 and txn2, we could use `execute` instead of `execute_async` to avoid the later `wait` step.

It's generally a good idea to avoid unnecessary detail / steps / etc in tests: keep the focus on what is intended to be tested in this case, and use the simplest version for everything else.

test-fixtures/src/bin/cache.rs
```diff
@@ -1202,6 +1203,72 @@ fn test_collapse_across_vary() {
...
+    assert!(pending_txn3.pending().unwrap(), "txn3 should be waiting on txn2");
+
+    let txn1 = pending_txn1.wait().unwrap();
+    let txn2 = pending_txn2.wait().unwrap();
+    assert!(txn1.must_insert_or_update());
+    assert!(txn2.must_insert_or_update());
+
+    // Txn1 completes with an empty vary rule (no .vary_by() on the insert).
+    // This pushes VaryRule::default() to the front of the key's vary_rules and
+    // inserts under the universal variant {}, which matches any request.
```
cceckman-at-fastly: A test should not describe the implementation; [a test should accept any correct implementation](https://cceckman.com/writing/pass-correct-code/). This portion of the comment should be removed.

Here it doesn't matter _how_ the rules are modified, or how we're tracking them. What matters is: "A result with an empty variant can match any request.")

test-fixtures/src/bin/cache.rs
```diff
@@ -1202,6 +1203,72 @@ fn test_collapse_across_vary() {
...
+    let mut writer = txn1.insert(Duration::from_secs(120)).execute().unwrap();
+    writer.write_all(b"the-response").unwrap();
+    writer.finish().unwrap();
+
+    // Txn2 is abandoned after Txn1 completes (matches the issue's ordering).
```
cceckman-at-fastly: "matches issue's ordering" doesn't matter so much :)

What's the underlying reason why we expect this result / why we need to cancel txn2 later? What do we expect to happen if txn2 is abandoned first, and why?

In fact, here's a thought experiment: if we didn't cancel txn2 here, what would we expect the result for txn3 to be?

cceckman-at-fastly (commented):
Overall looks good! A couple cleanup suggestions & one point for discussion, but overall this seems like it does the trick!

cceckman-at-fastly (approved):

PR #630, opened 2026-06-02
fix: snapshot CacheData on lookup to lock Found metadata (#492)

### LINKED ISSUE #492: Cache: metadata should not change for a Found
Found via testing in Compute-

Viceroy is written in such a way that metadata associated with a `Found` can change; namely, `Found` can get a known length after the fact (see `poll_known_length` in the test suite).

This can't actually happen in Compute -- the same `poll_known_length` function either immediately gets a length, or fails to ever get a length.

The Viceroy implementation should work the same way - metadata, including length, is locked-in at lookup time.

THad04:
Fixes #492.

Problem:
A found could acquire a known length after the fact. This does not match Compute, where 'poll_known_length' either gets a length on the first try or never gets one. Length should be locked in at lookup time.

Fix:
Snapshot the length when Found is constructed and have Found::length() return that snapshot. After this change, a Found whose lookup occurs midstream should report none forever, matching the behavior of Compute.

Added a test titled test_length_locked_up_at_lookup to the cache test fixtures. This test
1. Inserts without a known length and starts streaming
2. Looks up while the body is still streaming and asserts known_length() is None
3. Finishes the stream
4. Asserts that a fresh lookup can now see the final length
This test you should have failed prior to this fix.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -252,21 +252,13 @@ pub struct Found {
     /// We mirror the BodyHandle here when we create it; we can later check whether the handle is
     /// still valid, to find an outstanding read.
     pub last_body_handle: Option<BodyHandle>,
-
-    // Length is snapshoted at lookup time.
-    // CacheData::length() consults the live streamed body, which can transition from none to
-    // some as the body completes. This cannot happen on Compute. length is either known at lookup or
-    // never known. We capture it once here so Found::length() returns a stable value when it is called.
-    known_length: Option<u64>,
 }
 
 impl From<CacheData> for Found {
     fn from(data: CacheData) -> Self {
-        let known_length = data.length();
         Found {
             data,
             last_body_handle: None,
-            known_length,
         }
     }
 }
@@ -283,7 +275,7 @@ impl Found {
 
     /// The length of the cached object, if known at lookup time.
     pub fn length(&self) -> Option<u64> {
-        self.known_length
+        self.data.get_meta().length()
     }
 }
 
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -252,13 +252,22 @@ pub struct Found {
     /// We mirror the BodyHandle here when we create it; we can later check whether the handle is
     /// still valid, to find an outstanding read.
     pub last_body_handle: Option<BodyHandle>,
+
+    /// The length of the cached object, snapshotted at lookup time.
+    ///
+    /// On Compute, length is locked-in at lookup time: it is either known immediately or never
+    /// known. Viceroy mirrors this behavior by capturing the length once when `Found` is created,
+    /// rather than re-querying the (potentially still-streaming) body on each call.
+    length: Option<u64>,
 }
 
 impl From<Arc<CacheData>> for Found {
     fn from(data: Arc<CacheData>) -> Self {
+        let length = data.length();
         Found {
             data,
             last_body_handle: None,
+            length,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/cache/store.rs
```diff
@@ -500,7 +518,7 @@ impl Obligation {
         // Mild optimization: avoid re-acquiring the lock when we drop.
         // We've already cleared the obligation flag.
         self.completed = true;
-        data.into()
+        (*data).clone().into()
```
cceckman-at-fastly: What if `self.object.insert` returns a `CacheData`, instead of an `Arc<CacheData>`?

We want to keep the scope where we work with the `Arc<CacheData>` very small, as that will always present _an_ opportunity for a race. (...I think there might still be a very subtle one here!)

src/cache/store.rs
```diff
@@ -556,6 +574,23 @@ pub(crate) struct CacheData {
     body: CollectingBody,
 }
 
+impl Clone for CacheData {
+    fn clone(&self) -> Self {
+        // Resolve the length question at clone time.
+        // After this, the cloned CacheData's meta.length will be the single source of truth
```
cceckman-at-fastly: This is good! But the API winds up being somewhat confused -- `meta.length` is accessible, but so is `CacheData::length`, which will re-resolve (and may update).

Instead, let's make `CacheData::length` private, so it's only used at this point: when we make the copy. Then we know the length we get from the `Found` won't change, because it'll be the `meta`.

src/cache.rs
```diff
@@ -252,30 +252,38 @@ pub struct Found {
     /// We mirror the BodyHandle here when we create it; we can later check whether the handle is
     /// still valid, to find an outstanding read.
     pub last_body_handle: Option<BodyHandle>,
+
+    // Length is snapshoted at lookup time.
+    // CacheData::length() consults the live streamed body, which can transition from none to
+    // some as the body completes. This cannot happen on Compute. length is either known at lookup or
+    // never known. We capture it once here so Found::length() returns a stable value when it is called.
+    known_length: Option<u64>,
```
cceckman-at-fastly: If we have `data.meta().length`, why do we need this too?

src/cache.rs
```diff
@@ -252,30 +252,38 @@ pub struct Found {
...
 }
 
-impl From<Arc<CacheData>> for Found {
-    fn from(data: Arc<CacheData>) -> Self {
+impl From<CacheData> for Found {
+    fn from(data: CacheData) -> Self {
+        let known_length = data.length();
```
cceckman-at-fastly: This is a _separate_ snapshot, distinct from the one in `meta` -- taken at a different point in time.

Which is the right one?

src/cache.rs
```diff
@@ -337,7 +342,7 @@ impl Cache {
             .await;
         CacheEntry {
             key: key.clone(),
-            found: found.map(|v| v.into()),
+            found: found.map(|v| Found::from((*v).clone())),
```
cceckman-at-fastly, line 337: What does `.clone` do here?

THad04: .clone() is taking the snapshot. v is Arc<CacheData> from the map, *v de-references to CacheData, and .clone() makes an owned copy with length baked into meta at that moment.

cceckman-at-fastly: ah ok -- the `*v` was confusing me.

Sure -- `v.clone()` resolves to `Arc::clone(&v)`, `(*v).clone()` becomes `CacheData::clone(v)`.

cceckman-at-fastly (approved):
Nice!

PR #632, opened 2026-06-08
Add 'release-lto-stripped' profile

kpfleming:
This profile is used by the 'release' workflow, and provides two benefits:

1. There is no need to 'strip' symbols in the workflow, because the symbols will be stripped during the build.

2. 'fat' LTO mode is used to reduce the size of the binary and improve inlining and other optimization opportunities. With Rust 1.95.0 on x86_64 this reduces the size of the final binary from 41MB to 35MB.

Also adds a 'test-crates-lto' target to the Makefile to run the test suite against an LTO-optimized build, and extends the 'test' workflow to execute it along with the other test jobs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,23 +5,25 @@ on:
     tags:
       - "v*.*.*"
 
+  workflow_dispatch:
+
 jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-24.04, macos-14, windows-2022]
+        runner: [ubuntu24.04-8cpu-32ram, macos-14, windows-2022]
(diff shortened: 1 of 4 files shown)


kpfleming:
The `ubuntu-latest` runner may not have enough RAM to complete the `lto-test` job...

kpfleming:
Leaving in 'draft' until we can figure out the runner situation.

kailan (dismissed):
Looks good. Hopefully we can sort the runner situation.

kpfleming:
We now have a runner with 32GB of RAM which is able to complete the `lto-test` job.

kailan (approved):

PR #635, opened 2026-06-10
Add support for Healthchecking Dynamic Backends

camilee03:
Meant to support the changes made in [this](https://github.com/fastly/fst-compute-sdk-rs/pull/151) PR in the Rust SDK to support healthchecking hostcalls for dynamic backends

There is additional healthchecking-related terminology like [here](https://github.com/fastly/Viceroy/blob/40d549c87e5452b0fa6d839d7c047b05d0002da2/src/wiggle_abi/req_impl.rs#L607), which sets backend health to `BackendHealth::Unknown` (which the static backend sets as a config). This should be fine to leave as is.


dgryski:
Does this prevent sending http requests to unhealthy backends too?

camilee03:
@dgryski No, it doesn't. This is focused on adding healthchecks as an option to dynamic backends on creation re the work in ExecuteD along similar lines, not on the other side of when a user tries to query a created backend and healthcheck it.

Unless I've misinterpreted the code for Viceroy, and it's different from ExecuteD, I believe that functionality shouldn't ~need to be changed because once the dynamic backend is built, it is processed the same way as a static backend (lmk if that assumption is incorrect)?

dgryski:
When backend health was added to viceroy, the status was not checked during an http send.  So, no change then.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fix adapter codegen to avoid memory shift overflow, and support SIMD instructions. ([#629](https://github.com/fastly/Viceroy/pull/629))
 - Add support for sending `PendingRequestHandle` downstream ([#631](https://github.com/fastly/Viceroy/pull/631))
+- Add support for healthchecking dynamic backends ([#635](https://github.com/fastly/Viceroy/pull/635))
```
cceckman-at-fastly, line 4:
```suggestion
- Add support for healthchecking dynamic backends ([#635](https://github.com/fastly/Viceroy/pull/635))
- Always use an `unsafe` block to `unsafe_main_ptr`. ([#643](https://github.com/fastly/Viceroy/pull/643))

## 0.19.0 (2026-06-16)

- Fix adapter codegen to avoid memory shift overflow, and support SIMD instructions. ([#629](https://github.com/fastly/Viceroy/pull/629))
- Add support for sending `PendingRequestHandle` downstream ([#631](https://github.com/fastly/Viceroy/pull/631))
```

cceckman-at-fastly (dismissed):

cceckman-at-fastly (approved):

PR #636, opened 2026-06-11
Make sure we enable TLS for wss in addition to https

harmony7:
This is a quick follow-up for #621 

This PR corrects a bug to handling during WebSocket passthrough so that backends defined with the `wss://` schema are treated with TLS.

We already have code that handles https, this fix adds wss to the check.

### Background

When using Viceroy for WebSocket passthrough, it's reasonable to accept backends with the `ws://` or `wss://` schemas, so that one may write:

```toml
[local_server.backends.backend]
url = "wss://echo.websocket.org/"
override_host = "echo.websocket.org"
```

instead of:
```toml
[local_server.backends.backend]
url = "https://echo.websocket.org/"
override_host = "echo.websocket.org"
```

There is already code that adds a default port number `:443` to `wss` URLs to backends when used with WebSocket passthrough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Add support for sending `PendingRequestHandle` downstream ([#631](https://github.com/fastly/Viceroy/pull/631))
+- Fix to correctly apply TLS when using `wss:` URL schema with WebSockets passthrough ([#636](https://github.com/fastly/Viceroy/pull/636))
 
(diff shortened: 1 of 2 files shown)


kpfleming (dismissed):

kpfleming:
This should be fine, with the addition of a changelog entry.

harmony7:
I added a changelog entry, looks like I need a review again

kpfleming (approved):

PR #642, opened 2026-06-16
Update to Wasmtime 46.

dgohman-fastly:
The biggest source of diffs is the change from `anyhow::Error` to `wasmtime::Error`. I enabled the "anyhow" feature so that `wasmtime::Error` converts into `anyhow::Error`, which helps, but we still needed a bunch of updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Always use an `unsafe` block to `unsafe_main_ptr`. ([#643](https://github.com/fastly/Viceroy/pull/643))
+- Update to Wasmtime 45. ([#642](https://github.com/fastly/Viceroy/pull/642))
 
 ## 0.19.0 (2026-06-16)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,11 +138,11 @@ futures = "0.3.24"
 url = "2.3.1"
 
 # Wasmtime dependencies
-wasmtime = { version =  "39.0.2", features = ["call-hook"] }
-wasmtime-wasi = "39.0.2"
-wasmtime-wasi-io = "39.0.2"
-wasmtime-wasi-nn = "39.0.2"
-wiggle = "39.0.2"
+wasmtime = { version =  "45.0.1", features = ["anyhow", "call-hook"] }
+wasmtime-wasi = "45.0.1"
+wasmtime-wasi-io = "45.0.1"
+wasmtime-wasi-nn = "45.0.1"
+wiggle = "45.0.1"
 wat = "1.250.0"
 wasmparser = "0.250.0"
 wasm-encoder = { version = "0.250.0", features = ["wasmparser"] }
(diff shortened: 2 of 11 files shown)


src/error.rs
```diff
@@ -194,6 +194,12 @@ pub enum Error {
     BetweenBytesTimeout,
 }
 
+impl From<wasmtime::Error> for Error {
```
WhiteAbeLincoln: Is this impl necessary? I tried removing it and we only errored in linker setup (src/linking.rs) and the engine construction in (src/execute.rs), both could probably be fixed by a helper function or a bunch of `map_err` if we don't mind the noise. My worry is that this makes it easy to accidentally use the conversion to `Error::Other` when a variant with more information would be appropriate

src/execute.rs
```diff
@@ -634,7 +634,7 @@ impl ExecuteCtx {
         incoming_req: Request<hyper::Body>,
         local: SocketAddr,
         remote: SocketAddr,
-    ) -> Result<Response<Body>, Error> {
+    ) -> Result<Response<Body>, wasmtime::Error> {
```
WhiteAbeLincoln: Did this return type need to change? `handle_request` returns an `Error` and there aren't any other calls which produce a `wasmtime::Error`. It also places an unwrapped `wasmtime::Error` in our public interface which we probably don't want.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Update test-fixtures to fastly 0.13.1. ([#707](https://github.com/fastly/Viceroy/pull/707))
 - Internal unification of native/guest profiling config. (#[597](https://github.com/fastly/Viceroy/pull/597))
 - Remove the warning about Wasm Component support. ([#709](https://github.com/fastly/Viceroy/pull/709))
+- Update to Wasmtime 46. ([#642](https://github.com/fastly/Viceroy/pull/642))
```
WhiteAbeLincoln: This needs to move to the Unreleased section

Cargo.toml
WhiteAbeLincoln: I think this is a breaking change, since we use some types from the bumped crates in public signatures: `wasmtime::WasmFeatures` and `wasmtime::ProfilingStrategy`. Plus there's that change to the return type of `handle_request_with_runtime_error` (which we should probably revert).

I think if a user of `viceroy-lib` has their own dependency on `wasmtime` then an update of viceroy would break since rust starts treating the dependencies as distinct (despite the actual type definitions not having changed).

WhiteAbeLincoln (changes requested):
Looks pretty good. I added some comments. At minimum we should fix the changelog and probably the crate version.

src/error.rs
dgohman-fastly: I've now removed that impl, and added `map_err` calls, with a helper function in linker.rs to reduce noise.

WhiteAbeLincoln (approved):

PR #645, opened 2026-06-22
Actions workflows overhaul

### LINKED ISSUE #247: Prevent duplicate release drafts
Our github actions currently have a race condition that can cause two separate draft releases to get created if two build jobs finish at the same time. This happened for the 0.4.4 release:
image

This resulted in some artifacts not actually getting uploaded to the correct release.

I think we can use Github's [concurrency mechanism](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) at the job level to ensure that a draft release already exists before any of the build jobs get to [this step](https://github.com/itsrainy/Viceroy/blob/main/.github/workflows/release.yml#L78-L86).

kpfleming:
This is an almost complete overhaul of the Actions workflows.

1. Apply Zizmor and add a workflow which runs it against PRs. This
will ensure that various security properties of the workflows are
maintained. One significant change is that references to third-party
Actions are now pinned by commit hash.

2. Support optional repository-level variables to provide the names of
runners to be used in various workflows (except the 'release'
workflow). This allows private forks to specify their own runner
labels without having to modify the workflow files.

3. Use the 'gh' command line tool in the 'release' workflow instead of
the softprops/action-gh-release Action.

4. Use newer runners where available (macos-26 instead of macos-14,
windows-2025 instead of windows-2022).

5. Eliminate cross-compilation in 'release' workflow by using ARM64
runners for producing ARM64 binaries.

6. Remove duplication and unnecessary attributes in various steps.

7. Enable Dependabot version updates for Cargo and GitHub Actions
ecosystems.


kpfleming:
After this is merged the list of 'required checks' in the branch protection settings will need to be updated. The `release` workflow will also need to be manually tested, which will test the release-creation and build jobs but will not test the release-publication and `cargo publish` jobs - those won't be tested until the next actual release is made.

.github/workflows/release.yml
```diff
@@ -7,55 +7,86 @@ on:
...
+  run:
+    shell: bash
+
 jobs:
+  prep-release:
+    runs-on: ${{ case(vars.LINUX_RUNNER_SMALL != '', vars.LINUX_RUNNER_SMALL, 'ubuntu-slim') }}
+    permissions:
+      contents: write
+    steps:
+      - name: Create draft release
```
kailan, line 20: I'm pleased about this explicit creation as sometimes we'd run into race conditions in the past where two builds complete at the same time and each create their own release

.github/workflows/release.yml
```diff
@@ -7,55 +7,86 @@ on:
 
   workflow_dispatch:
```
kailan, line 8: Unrelated to your changes but I can't see a situation in which workflow_dispatch is necessary for this workflow

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,28 @@
+name: GitHub Actions Security Analysis with zizmor
```
kailan, line 1: Not a blocker but this file is `.yaml` while the rest are `.yml`

.github/workflows/release.yml
```diff
@@ -7,55 +7,86 @@ on:
...
+          - runner: windows-2025
             name: windows
             rust_abi: pc-windows-msvc
             extension: .exe
+
           - arch: arm64
             rust_arch: aarch64
+
           - arch: amd64
             rust_arch: x86_64
```
kailan: This section feels a little tricky to understand. Can I suggest we just explicitly define the runner configurations instead?

```suggestion
        include:
          - runner: ubuntu-24.04
            arch: amd64
            rust_arch: x86_64
            rust_abi: unknown-linux-gnu
            name: linux

          - runner: ubuntu-24.04-arm
            arch: arm64
            rust_arch: aarch64
            rust_abi: unknown-linux-gnu
            name: linux

          - runner: macos-26
            arch: arm64
            rust_arch: aarch64
            rust_abi: apple-darwin
            name: darwin

          - runner: macos-26-intel
            arch: amd64
            rust_arch: x86_64
            rust_abi: apple-darwin
            name: darwin

          - runner: windows-2025
            arch: amd64
            rust_arch: x86_64
            rust_abi: pc-windows-msvc
            name: windows
            extension: .exe
```

.github/workflows/release.yml
kpfleming: It's not necessary, but a couple of times there's been a desire to make a binary build of a branch with some candidate changes so a customer could test them out, and this will allow that to happen. A maintainer could use workflow_dispatch to make a 'release' from their branch and then either grab the binaries from it or mark the release as 'prerelease' and publish it so the customer could grab the binaries directly.

.github/workflows/zizmor.yml
kpfleming: Ha! The opposite problem that customer-documentation had :-)

.github/workflows/release.yml
kpfleming: Yes, this is 1000 times better since the matrix is so sparse. I just followed the existing model but should have re-thought it.

.github/workflows/zizmor.yml
kpfleming: Addressed in 3545de5.

.github/workflows/release.yml
kpfleming: Addressed in 3545de5.

kpfleming: (GitHub failed repeatedly trying to 'Apply suggestion' so I did it locally and pushed)

kailan (dismissed):

kailan:
This PR resolves #247.

.github/workflows/test.yml
```diff
@@ -1,38 +1,53 @@
...
+
 jobs:
   test:
     strategy:
       matrix:
-        platform: [ubuntu-22.04, ubuntu-24.04, windows-2022, macos-14, macos-15]
+        runner:
+          - ${{ case(vars.LINUX_RUNNER_ARM64 != '', vars.LINUX_RUNNER_ARM64, 'ubuntu-24.04-arm') }}
+          - ${{ case(vars.MACOS_RUNNER_ARM64 != '', vars.MACOS_RUNNER_ARM64, 'macos-26') }}
+          - ${{ case(vars.WINDOWS_RUNNER != '', vars.WINDOWS_RUNNER, 'windows-2025') }}
```
cceckman-at-fastly, line 25: What if we don't have any such runners / want to skip these runs? We still get them, using the public runners?

.github/workflows/test.yml
```diff
@@ -42,58 +57,67 @@ jobs:
...
       run: make trap-test-ci
-      shell: bash
 
   # Run the LTO test in an isolated job. It uses a large amount of RAM
   # and the differences between it and the regular build are not
-  # platform-dependent.
+  # runner-dependent.
   lto-test:
-    runs-on: ubuntu24.04-8cpu-32ram
+    runs-on: ${{ case(vars.LINUX_RUNNER_LARGE_AMD64 != '', vars.LINUX_RUNNER_LARGE_AMD64, 'ubuntu24.04-8cpu-32ram') }}
```
cceckman-at-fastly, line 95: Looks like this is the only linux-amd64 test set we run?

Which looks like it still covers all the other tests except the trap test.

.github/workflows/test.yml
```diff
@@ -42,58 +57,67 @@ jobs:
   trap-test:
     strategy:
       matrix:
-        platform: [ubuntu-22.04, ubuntu-24.04, windows-2022, macos-14, macos-15]
-    runs-on: ${{ matrix.platform }}
+        runner:
+          - ${{ case(vars.LINUX_RUNNER_ARM64 != '', vars.LINUX_RUNNER_ARM64, 'ubuntu-24.04-arm') }}
+          - ${{ case(vars.MACOS_RUNNER_ARM64 != '', vars.MACOS_RUNNER_ARM64, 'macos-26') }}
+          - ${{ case(vars.WINDOWS_RUNNER != '', vars.WINDOWS_RUNNER, 'windows-2025') }}
+    runs-on: ${{ matrix.runner }}
```
cceckman-at-fastly, line 68: We should not stop running the trap-test on linux-AMD64.

.github/workflows/test.yml
kpfleming: Yes, the public runner labels are the defaults.

.github/workflows/test.yml
kpfleming: I can revert the Linux tests to use AMD64 instead of ARM64 if that's important. I generally use ARM64 when I can because the runners are faster and less expensive, but sometimes specific architecture usage is required. I don't think we need to run these tests on *two* Linux architectures, one will be sufficient.

.github/workflows/test.yml
kpfleming: GitHub Actions doesn't provide a way to skip certain tests in a matrix based on user input as far as I know, so the fork repository is stuck running the same tests :-)

.github/workflows/test.yml
cceckman-at-fastly: There's a nonzero amount of Wasmtime code that is architecture-specific; we avoid making any such assertions in Viceroy, but that doesn't mean we're free from the changes.

I don't think we should offer a release for a triple unless we have run the tests on that triple.

cceckman-at-fastly: (I don't mean to claim "every merge needs to run all the tests on all platforms", but I do think  "our release flow needs to run the tests for a platform before publishing for that platform")

kpfleming: Agreed, but the simplest solution for that now is to just run the tests on all the platforms that will be used for releases.

kpfleming: Addressed in f8b2647.

.github/workflows/test.yml
kpfleming: Addressed in f8b2647.

cceckman-at-fastly (approved):

kailan (approved):

kpfleming:
Branch protection settings have been updated. I'll create a branch to test the release workflow.

PR #657, opened 2026-07-03
Implement the blocking KV store WIT functions.

dgohman-fastly:
Implement the KV store `list`/`lookup`/`insert`/`delete` functions, which are like their `*_async` coutnerparts, except that they wait for the result.


dgohman-fastly:
Replaced by #702.

PR #658, opened 2026-07-07
Add support for headers-based bot detection, plus HeadlessBot support

zkat:


dgryski (dismissed):
LGTM.  Wouldn't mind a look over by someone with more rust and viceroy experience though.

dmartin (commented):
I'm a bit concerned about allowing the bot category string (`X-Fastly-Bot-Category`) and kind value (`X-Fastly-Bot-Category-Kind`) to be customized independently, because:
- there would always be a 1-to-1 mapping in production
- the `kind` value is more important, since the SDKs parse it into an enum value, but I don't think we currently do or want to publish those int-to-string mappings.

Maybe the single header `X-Fastly-Bot-Category` could be reverse-mapped to its corresponding `kind` value, defaulting to `u32::MAX` if there is no string reverse match (which will be translated to an "Unknown" kind by the SDKs)

zkat:
Part of the reason I did it this way is for folks to be able to independently test arbitrary values, since the documentation for `bot_category` explicitly mentions that these category strings can change arbitrarily and should only be for reporting, whereas the `kind` is stable and reliable.

dmartin:
Fair, allowing different string values makes sense. Perhaps we can just list out the `kind` mappings wherever the `X-Fastly-Bot-Category-Kind` header ends up being documented.

zkat:
I went ahead and just made it a single header. I think I was being too fancy. What matters in the end is that you be able to give arbitrary enum values (through `u32`s), and so I've added that possibility. lmk what you think!

wasm_abi/wit/deps/fastly/compute.wit
```diff
@@ -413,6 +413,8 @@ interface http-downstream {
     search-engine-optimization,
     /// Security analysis tools that inspect your website for vulnerabilities, misconfigurations and other security features.
     security-tools,
+    /// A headless bot.
+    headless,
```
dgohman-fastly: Unfortunately, adding a value to an enum is currently backwards-incompatible in WIT, so we can't make this change. The new value will need to go through the `extra-bot-category` resource.

src/component/compute/http_downstream.rs
```diff
@@ -223,46 +223,127 @@ impl http_downstream::Host for ComponentCtx {
...
+                "content-fetcher" | "5" => ContentFetcher,
+                "monitoring-site-tools" | "6" => MonitoringSiteTools,
+                "online-marketing" | "7" => OnlineMarketing,
+                "page-preview" | "8" => PagePreview,
+                "platform-integration" | "9" => PlatformIntegrations,
+                "research" | "10" => Research,
+                "search-engine-crawler" | "11" => SearchEngineCrawler,
+                "search-engine-optimization" | "12" => SearchEngineOptimization,
+                "security-tools" | "13" => SecurityTools,
+                "headless" | "14" => Headless,
```
dmartin: In production, a detected headless bot will return `Extra`, with an inner handle to an `ExtraBotCategory { raw: 14 }`. We should probably match that behavior here rather than having the Viceroy-bundled WIT diverge. (Ah, looks like Dan beat me to it above)

dmartin (changes requested):
One new note, the new handling of the category header otherwise LGTM though!

wasm_abi/wit/deps/fastly/compute.wit
zkat: dang ok. Thanks.

zkat: fixed

dmartin: It looks like the implementation now uses the extra resource, but this variant should be removed from the WIT definition as well.

dmartin (changes requested):

wasm_abi/wit/deps/fastly/compute.wit
zkat: huh. I thought I did that already. Thanks!

dmartin (approved):

PR #662, opened 2026-07-09
v0.20.0

zkat:


kpfleming (dismissed):
Ideally the code changes (adding `let _` to absorb function results) wouldn't be in a 'release' PR, but they aren't significant enough to be a concern.

kpfleming (approved):

PR #664, opened 2026-07-14
Properly apply ValueAbsent in wiggle_abi for bot detection

zkat:
This needs wiggle-level tests, but I don't know how to write those? I'll keep looking around, but getting pointed at a test that checks both the regular api and the wiggle api would be great.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix wiggle abi implementation for bot detection mocking. ([#664](https://github.com/fastly/Viceroy/pull/664))
+
 ## 0.20.0 (2026-07-09)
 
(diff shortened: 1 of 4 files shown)


dmartin (dismissed):
The `"" -> ValueAbsent` changes LGTM. I usually haven't seen ABI-level (in addition to SDK-level) tests when reviewing PRs for this repo... but if you wanted to test the wiggle ABI directly, you could add a test fixture that directly calls the `fastly_http_downstream::downstream_bot*` methods via `fastly-sys`.

zkat:
@dmartin ah thank you! I've added tests for the wiggle abi directly, then :) (I'll need another +1)

dgryski (approved):
I believe you.

PR #671, opened 2026-07-30
Add a github workflow for publishing the WIT API to a GHCR repository.

dgohman-fastly:


kpfleming:
No need for a changelog entry, this is an internal tooling change and isn't relevant for users of Viceroy.

.github/workflows/publish-wit.yml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      packages: write
+      contents: write
```
kpfleming: I don't see any steps in the workflow which write to the repository itself, only the package repository and attestations. Do you know which step which requires this permission?

dgohman-fastly: None of the steps do; that was an oversight on my part. I've now removed it.

kpfleming (approved):
Looks good to me. The WKG parts I don't know about, but the workflow itself is well-written and should be secure.

.github/workflows/publish-wit.yml
```diff
@@ -0,0 +1,62 @@
...
+      id-token: write
+      packages: write
+      attestations: write
+
+    env:
+      WKG_VERSION: 0.15.0
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
erikrose, line 21: Thank you for hashing. :-D

.github/workflows/publish-wit.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Publish to GitHub Container Registry
+        id: publish
+        uses: bytecodealliance/wkg-github-action@10b3b04b9059ba46208cd7daf7d352af14bded0f # v5
+        with:
+          oci-reference-without-tag: 'ghcr.io/fastly/Viceroy/fastly-compute'
+          file: 'fastly-compute.wasm'
+          description: 'The APIs of the Fastly Compute platform'
+          source: 'https://github.com/fastly/Viceroy'
+          homepage: 'https://www.fastly.com/documentation/guides/compute/'
+          version: '0.0.0'
```
erikrose, line 53: Did you intend to make this static? As it stands, the latest 0.0.0 push will overwrite any former; GHCR always has mutable tags.

erikrose (commented):
Aside from my question on intended tag overwriting, looks good!

.github/workflows/publish-wit.yml
kpfleming: It's required in this repository :-)

.github/workflows/publish-wit.yml
dgohman-fastly: Yeah; this is all still experimental and only run from a workflow_dispatch for now. In the future, we should set it up to auto-publish when there's a Viceroy release, and make it pick up the appropriate version number.

.github/workflows/publish-wit.yml
erikrose: Even better! :-D

PR #672, opened 2026-07-30
Allow shared options before explicit subcommands

### LINKED ISSUE #358: CLI options are mutally exclusive with `run` mode
The one-shot `run` mode doesn't allow any CLI options, but `serve` (the default) does. CLI options should be handled for both.

Pybsama:
## Summary

- allow shared serve/run options before explicit `serve` and `run` subcommands
- preserve clap duplicate, count, append, command-specific, and trailing guest-argument semantics
- keep the default serve form and make root help describe the valid serve, run, and adapt combinations

## Testing

- `cargo test -p viceroy opts_tests --no-fail-fast`
- `make ci`
- `make trap-test-ci`
- `cargo clippy -p viceroy --all-targets --all-features --no-deps -- -D warnings`
- `cd wasm_abi/adapter && cargo build --release -p viceroy-component-adapter --target wasm32-unknown-unknown`

The workspace-wide Clippy target currently reports six warnings in unchanged
`viceroy-lib` files; the changed CLI package passes strict Clippy on the pinned
Rust 1.95 toolchain.

Fixes #358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Allow shared CLI options before explicit `serve` and `run` subcommands. ([#358](https://github.com/fastly/Viceroy/issues/358))
+
 ## 0.20.1 (2026-07-16)
 
 - Fix wiggle abi implementation for bot detection mocking. ([#664](https://github.com/fastly/Viceroy/pull/664))
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -30,8 +30,9 @@ use {
 #[tokio::main]
 async fn main() -> ExitCode {
     // Parse the command-line options, exiting if there are any errors
-    let opts = Opts::parse();
-    let cmd = opts.command.unwrap_or(Commands::Serve(opts.serve));
+    let cmd = Opts::parse()
+        .into_command()
+        .unwrap_or_else(|error| error.exit());
     match cmd {
         Commands::Run(run_args) => subcommands::run::exec(run_args).await,
         Commands::Serve(serve_args) => subcommands::serve::exec(serve_args).await,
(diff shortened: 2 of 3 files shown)


Pybsama:
Revalidated this draft against current `main` (`4ba94a25f8ed4d4b26a2e1ce85ccec9fff1936e0`). The code changes transplant cleanly; the only textual conflict is the changelog section created by the v0.21.0 release.

Current `main` also added `--enable-local-websocket-passthrough`. A direct transplant exposed three deterministic regressions in the option-resolution layer: a prefix `=false` value was dropped, the same option on both sides of `run` was not rejected, and the option was accepted before `adapt`. The new tests produced 28 passes / 3 failures before the compatibility adjustment and 31 / 31 afterward. The adjustment preserves the public default of `true`, distinguishes omission from an explicit value, merges a prefix value, rejects duplicates, and keeps the option out of `adapt`.

Validation with the repository-pinned Rust 1.95 toolchain:

- CLI option tests: 31 / 31, plus 100 consecutive repetitions
- integration tests: 176 / 176
- `viceroy-lib`: 125 / 125; doctests: 2 / 2
- trap tests: 2 / 2
- component-adapter release build: passed
- changed-package Clippy with `-D warnings`, formatting, diff check, and the 154-file `viceroy-lib` publish dry run: passed

The full-workspace Clippy command still reports the same six warnings in unchanged `viceroy-lib` files; none are in this PR's files.

The resolved local commit is `bbe7fb5953ef726baa106df0d5ef3d2c97549c49` with tree `c48d8eb566b80fb751556b010bf2b5e15a2e12da`. I attempted an exact force-with-lease update from the unchanged public head, but GitHub rejected it before changing the branch because current `main` contains the new `.github/workflows/publish-wit.yml` and the existing OAuth token lacks `workflow` scope. I did not widen credentials. The public head therefore remains `a35f0a3a4f82e9233b33bcbd177cc7a295743488`, and the PR remains Draft until the branch can be updated safely.

PR #673, opened 2026-07-30
Make the OCI package version patch the WIT package version.

dgohman-fastly:
fastly:compute is currently at version 0.1.0.


.github/workflows/publish-wit.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           description: 'The APIs of the Fastly Compute platform'
           source: 'https://github.com/fastly/Viceroy'
           homepage: 'https://www.fastly.com/documentation/guides/compute/'
-          version: '0.0.0'
+          version: '0.1.0'
```
erikrose, line 53: LG. Commit message: "patch"→"match".

erikrose (approved):

PR #678, opened 2026-08-07
Correct the mechanism that reports when Fanout and WebSocket passthrough are `unsupported` to the guest

harmony7:
Fixes two reported bugs that share a root cause: a guest can never observe that Fanout or WebSocket passthrough is disabled, because the failure happens after the guest has already returned.

## Problem

Calling `redirect_to_grip_proxy` without `--local-pushpin-proxy-port` succeeds from the guest's point of view. Viceroy then generates a plain-text 500 (`Pushpin handoff signaled, but Pushpin mode not enabled.`) that the guest played no part in producing:

```
HTTP/1.1 500 Internal Server Error
Pushpin handoff signaled, but Pushpin mode not enabled.
```

Both WIT functions are declared `result<_, error>`, and `compute.wit` documents them as usable only on services with the corresponding product enabled. SDKs are already written against that contract — the Rust SDK checks for `FastlyStatus::Unsupported` but Viceroy never produced it, so that branch was untestable locally.

WebSocket passthrough had no concept of a disabled state at all: a handoff to a valid backend always proxied through.

## Changes

**Fanout.** The enable signal (`--local-pushpin-proxy-port`) is unchanged; only the timing and shape of the failure. The check moves into `Sandbox::redirect_downstream_to_pushpin`, before the handoff is signaled, returning `Error::Unsupported`.

**WebSocket passthrough.** New `--enable-local-websocket-passthrough=<true|false>`, default `true` so existing behavior is preserved. When `false`, `Sandbox::redirect_downstream_to_backend` reports `unsupported` the same way. A companion Fastly CLI change will set this from `fastly.toml`.

Each guard sits in the single method that funnels every entry point for its hostcall, so one guard each covers the wiggle v1 and v2 paths and the component path. Both guards run *after* the existing already-sent check.

`Error::Unsupported` already maps to `FastlyStatus::Unsupported` and `types::Error::Unsupported`, so **no WIT, witx, or adapter changes were needed**.

Before: an opaque 500. After: the guest's own response, reporting `FastlyStatus::UNSUPPORTED` (5), identically in core-wasm and component modes.

## Testing

New `test-fixtures/src/bin/handoff.rs` reports the raw `FastlyStatus` it observes. `cli/tests/integration/handoff.rs` covers, via `viceroy_test!` so each runs as both core-wasm and component:

- Fanout disabled → guest observes `UNSUPPORTED`, and gets its own response rather than the old 500
- WebSocket passthrough disabled → guest observes `UNSUPPORTED`
- WebSocket passthrough enabled → still proxies to the backend
- Response already sent, then handoff on a disabled product → the already-sent error wins, locking in the ordering that matches ExecuteD

`make ci` (174 integration, 125 lib) and `make trap-test` pass. Note that the 6 `make clippy` errors on this branch are pre-existing — verified identical on a clean `main`, and clippy is not part of `make ci` or the CI workflow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   generated a plain-text 500 that the guest had no part in producing. The hostcall now returns
   `FastlyStatus::Unsupported` before the handoff is signaled, matching the behavior of a
   deployed service that does not have Fanout enabled, so that SDK-level error handling can run.
+  ([#678](https://github.com/fastly/Viceroy/pull/678))
 
 - Add `--enable-local-websocket-passthrough=<true|false>` (default `true`), to simulate a
   service that does not have the WebSockets feature enabled.
@@ -15,6 +16,7 @@
   When set to `false`, `redirect_to_websocket_proxy` (and `_v2`) reports
   `FastlyStatus::Unsupported` to the guest rather than proxying the request, which makes the
   "WebSockets not enabled" path testable locally for the first time.
+  ([#678](https://github.com/fastly/Viceroy/pull/678))
 
 ## 0.20.1 (2026-07-16)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 ## Unreleased
 
+- Report Fanout as `unsupported` to the guest when it is not enabled, instead of failing after
+  the guest has returned.
+
+  Previously, calling `redirect_to_grip_proxy` (or `_v2`) without
+  `--local-pushpin-proxy-port` succeeded from the guest's point of view, and Viceroy then
+  generated a plain-text 500 that the guest had no part in producing. The hostcall now returns
+  `FastlyStatus::Unsupported` before the handoff is signaled, matching the behavior of a
+  deployed service that does not have Fanout enabled, so that SDK-level error handling can run.
+
+- Add `--enable-local-websocket-passthrough=<true|false>` (default `true`), to simulate a
+  service that does not have the WebSockets feature enabled.
+
+  When set to `false`, `redirect_to_websocket_proxy` (and `_v2`) reports
+  `FastlyStatus::Unsupported` to the guest rather than proxying the request, which makes the
+  "WebSockets not enabled" path testable locally for the first time.
+
 ## 0.20.1 (2026-07-16)
 
 - Fix wiggle abi implementation for bot detection mocking. ([#664](https://github.com/fastly/Viceroy/pull/664))
--- a/src/sandbox.rs
+++ b/src/sandbox.rs
@@ -288,10 +288,10 @@ impl Sandbox {
         // state through this hostcall's `unsupported` result, so guests (and the SDKs built on
         // them) can only exercise that path if we do the same.
         if self.ctx.local_pushpin_proxy_port().is_none() {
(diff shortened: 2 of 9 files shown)


zkat (dismissed):

src/sandbox.rs
```diff
@@ -279,14 +279,30 @@ impl Sandbox {
...
+
+        // Report "not enabled" to the guest here, rather than letting the handoff proceed and
+        // failing once the guest has already returned. Deployed services signal the disabled
+        // state through this hostcall's `unsupported` result, so guests (and the SDKs built on
+        // them) can only exercise that path if we do the same.
+        if self.ctx.local_pushpin_proxy_port().is_none() {
+            return Err(Error::Unsupported {
+                msg: "Fanout is not enabled on this service \
+                      (run Viceroy with --local-pushpin-proxy-port to enable it)",
+            });
```
kailan, line 294: I don't think this message ever gets surfaced to the dev unless they explicitly opt into verbose logging

- default verbosity → server log completely empty (default filter is error only, cli/src/main.rs:49-54)
- -v → still nothing
- -vv → DEBUG ... Hostcall yielded an error: Unsupported operation: Fanout is not enabled on this service (run Viceroy with --local-pushpin-proxy-port to enable it) (via the catch-all at src/wiggle_abi.rs:248)
- -vv in component mode → nothing at all; From<error::Error> for types::Error (src/component/compute/error.rs:214) drops msg, and there is no logging on that path

It could be worth adding a log line to these guards so that developers won't be confused why forwarding isn't working.

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 ## Unreleased
 
+- Report Fanout as `unsupported` to the guest when it is not enabled, instead of failing after
```
kailan, line 3: Would be good to add a PR link here like the other entries

kailan (changes requested):
This looks great! I have a comment around UX and a minor changelog request

src/sandbox.rs
harmony7: Good catch, and the verbosity breakdown was helpful. I agree the disabled state was effectively invisible unless you knew to reach for `-vv`.

I've set it to `error!` so that the default filter catches it: this is a misconfiguration rather than normal operation, and it's at most one per request since a handoff is terminal.

Also placing the log right next to the `Err`, in `Sandbox`, so the single call site covers wiggle v1/v2 *and* the component path, which also sidesteps `From<error::Error> for types::Error` dropping `msg`, no change needed there.

CHANGELOG.md
harmony7: this is one of those where the number isn't determined til you open the PR but you can't open the PR until you commit the file --- but yeah i've added it now, thanks.

harmony7:
ready for re-review @kailan

harmony7:
unfortunately the push invalidated the approval from @zkat too..

kailan (dismissed):
looks great!

zkat (dismissed):

harmony7:
I just did a merge resolution only in the changelogs file and it dismissed the reviews. ~so I'm going to bypass requirements and merge.~ Evidently I can't do that so I will have to re-ask for reviews

kailan (approved):

PR #685, opened 2026-08-25
Upgrade the dependencies that don't require a migration

kailan:


src/upstream.rs
```diff
@@ -45,10 +45,18 @@ pub struct TlsConfig {
 
 impl TlsConfig {
     pub fn new() -> Result<TlsConfig, Error> {
-        let certs = rustls_native_certs::load_native_certs().map_err(Error::BadCerts)?;
+        let native_certs = rustls_native_certs::load_native_certs();
+        for err in &native_certs.errors {
+            warn!("could not load native certificates: {}", err);
```
kailan, line 50: This _is_ a logic change but I think it's acceptable.

kailan:
Looks like CI is failing after merging `main` in. Needs investigation.

WhiteAbeLincoln:
Replaced by #715

PR #697, opened 2026-09-08
Fix invalid safety comment in viceroy cli

WhiteAbeLincoln:
install_tracing_subscriber set `$RUST_LOG` using `std::env::set_env`. This is unsafe in multi-threaded programs. The safety comment claimed this was fine since it was called in main before threads were created, but it was incorrect due to `#[tokio::main]` which sets up worker threads.

This was unlikely to cause any issues in real usage (worker threads aren't active until `.await` is called), but it's a quick fix.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -39,26 +39,29 @@ async fn main() -> ExitCode {
     }
 }
 
-fn install_tracing_subscriber(verbosity: u8) {
-    // Default to whatever a user provides, but if not set logging to work for
-    // viceroy and viceroy-lib so that they can have output in the terminal
-    if env::var("RUST_LOG").ok().is_none() {
-        // SAFETY: We are called early in `main` when there are no other
-        // threads created yet.
-        unsafe {
-            match verbosity {
-                0 => env::set_var("RUST_LOG", "viceroy=error,viceroy-lib=error"),
-                1 => env::set_var("RUST_LOG", "viceroy=info,viceroy-lib=info"),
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -74,7 +77,75 @@ fn install_tracing_subscriber(verbosity: u8) {
     }
     event!(
         Level::DEBUG,
-        "RUST_LOG set to '{}'",
-        env::var("RUST_LOG").unwrap_or_else(|_| String::from("<Could not get env>"))
+        filter = %filter_str,
+        "log filter configured"
```
anthony-gomez-fastly, line 81: it looks like the old output returned the level, do we want to do that in the new one?

WhiteAbeLincoln: I think it just printed the value of $RUST_LOG? With this change we don't modify that variable anymore so we can't keep the log statement as is. Printing the value of filter_str should be equivalent since it'll be $RUST_LOG or our default as determined by `tracing_log_filter`. I can revert the wording to the prior version though, I changed it since it wasn't technically accurate.

anthony-gomez-fastly: ah i think i misread the rust then, I didn't get that `filter` would be read as the var and not the string literal

anthony-gomez-fastly (approved):
LGTM

cli/src/main.rs
WhiteAbeLincoln: Yeah the tracing macro syntax is interesting, I have to look it up whenever I try to do something. I think this one will add `filter` as a structured attribute on the log, and use the Display implementation to get the value from the `filter_str` variable.

PR #705, opened 2026-09-10
Allow debugging integration tests in editor

WhiteAbeLincoln:
Paths were dependent on current working directory, which broke debugging in Zed because it launched the process from within the target dir. Also adds configuration so rust-analyzer detects the extra crates.


anthony-gomez-fastly (approved):
LGTM

PR #706, opened 2026-09-14
Fix panic on pollable drop

### LINKED ISSUE #696: `cache.entry.step` returns a `pollable` that aliases the `entry`, so a guest which drops the `pollable` destroys the `entry`
**Repro:** https://github.com/fastly/devex-eng-repros/tree/viceroy-core-cache-repros/viceroy-core-cache-repros
— a Rust guest for `wasm32-wasip2` on the raw component bindings, no SDK, which is what lets it choose
exactly when the `pollable` is dropped. The routes for this issue are `/step-drop` and its control
`/step-forget`. Its README carries the transcripts and the mechanism in Viceroy's own source. The same
repro also covers a second, unrelated finding about an abandoned insert.

## Summary

With the Core Cache API, `cache.entry.step` returns an **owned** `pollable`:

```wit
/// Returns a `pollable` representing the next step of work for this `entry`. The `pollable` can be used to wait until this `entry` is unblocked.
///
/// The `entry` may require `step`s to complete. When this returns `none`, the entry is ready.
step: func() -> option<pollable>;
```

Viceroy returns the `entry`'s own handle value inside that `pollable`. The two resources then share
one slot in the async-item table, and they have **opposite** semantics on a drop. So a guest that
takes the `pollable`, waits on it, and lets it go out of scope destroys its own `entry`. **The next
call on that `entry` panics the Viceroy worker**, and the client gets no reply.

A guest cannot defend itself, because the failure arrives as a host panic rather than as an `error`
it could handle. The guest is also doing nothing wrong: the WIT gives it an owned resource, so it is
entitled to drop it.

**The aliasing itself is not the bug, and the WIT cannot avoid it.** `borrow<T>` is not allowed in a
return position. `wasm-tools` refuses `step: func() -> option<borrow<pollable>>` with "returns a type
which contains a `borrow<T>` which is not supported". An owned `pollable` is therefore the only thing
`step` can return. The bug is that a drop of it frees the shared slot.

**The platform aliases the same handle and treats such a `pollable` as a borrow when it is dropped.**
If the async item behind the handle is a cache entry, a replace entry or an HTTP cache entry, then a
drop releases the `pollable` only, and leaves the item in the table. Viceroy differs in that one
place.

## The mechanism, on `main`

`api::HostEntry::step` in `src/component/compute/cache.rs` hands back the entry's own handle:

```rust
fn step(
    &mut self,
    entry: Resource<api::Entry>,
) -> wasmtime::Result<Option<Resource<api::Pollable>>> {
    let cache_handle = CacheHandle::from(entry);
    let async_handle = cache_handle.into();
    let is_ready = self.sandbox_mut().async_item_mut(async_handle)?.is_ready();

    if is_ready {
        Ok(None)
    } else {
        Ok(Some(AsyncItemHandle::from(async_handle).into()))
    }
}
```

The two drops disagree. `api::HostEntry::drop` in the same file does nothing, which is correct,
because an `entry` belongs to `close-entry`:

```rust
fn drop(&mut self, _entry: Resource<api::Entry>) -> wasmtime::Result<()> {
    Ok(())
}
```

`async_io::HostPollable::drop` in `src/component/compute/async_io.rs` frees the slot:

```rust
fn drop(&mut self, handle: Resource<async_io::Pollable>) -> wasmtime::Result<()> {
    let handle = wiggle_abi::types::AsyncItemHandle::from(handle).into();

    // Use `.take_async_item` instead of manipulating
    // `self.sandbox_mut().async_items` directly, so that any extra state
    // associated with the item is also cleared.
    let _ = self.sandbox_mut().take_async_item(handle).unwrap();

    Ok(())
}
```

So the `pollable` drop takes the cache entry out of the table, and the `entry` handle then names an
empty slot.

## Steps to reproduce

Clone the repro linked above and run it from its `viceroy-core-cache-repros/` directory. **Run the
control first**, so that a panic in the second command is attributable to the drop and to nothing else.

```
fastly compute build
fastly compute serve --skip-build
curl "http://127.0.0.1:7676/step-forget?key=demo"   # the control: answers 200
curl "http://127.0.0.1:7676/step-drop?key=demo"     # panics the worker, so curl gets nothing
```

By hand, the sequence is:

1. Call `cache.entry.transaction-lookup`, and then `cache.entry.step` until it answers `some`.
   **No contention is needed.** `await-entry` re-wraps the resolved entry in a task it has just
   spawned, and that task is often incomplete when the guest asks, so an ordinary uncontended lookup
   reaches this. About one attempt in three answered `none` for us, so the repro retries.
2. Call `async-io.select` on the `pollable`, and wait for the signal.
3. **Drop the `pollable`.**
4. Call anything on the `entry`, such as `get-state`.

## What happens

Step 4 panics the Viceroy worker, and the client gets no reply at all:

```
thread 'tokio-rt-worker' panicked at src/component/compute/error.rs:38:9:
`HandleError` should not be used in components

thread 'tokio-rt-worker' panicked at src/execute.rs:755:14:
guest worker finished without panicking: JoinError::Panic(..., "`HandleError` should not be used in components", ...)
```

The freed slot yields a `HandleError`, and the component error mapping refuses that variant with a
`panic!`, so the failure arrives as a host panic rather than as an `error` the guest could handle.

**The control is what makes this a finding and not a guess.** `/step-forget` runs the identical
lookup, `step` and `select`, and then leaks the `pollable` with `mem::forget` so that no
`[resource-drop]` reaches the host. It answers 200, and `get-state` reports
`must-insert-or-update`. Only the drop differs.

## Environment

- **Viceroy 0.21.0**, from `fastly compute install-tools`.
- rustc 1.94.0, target `wasm32-wasip2`, `wit-bindgen` 0.57.1.
- `/step-drop` panicked on 5 of 5 runs. `/step-forget` answered 200 on 5 of 5.

## Scope

`cache.entry.step` is the only call that can show this today. `cache.replace-entry.step` and
`http-cache.entry.step` both answer `Unsupported`, so neither is reachable. Both deserve the same
treatment when that support arrives, and #692 adds the `replace` family.

## Why this has not broken more guests

A guest only meets this if it drops the `pollable`. The .NET SDK never does: it parks the handle in
its event loop and leaves it, so it leaks the handle, and the leak is what keeps its cache lookups
working. That is incidental rather than designed, and it is a poor thing for any SDK to depend on. A
guest which writes the obvious `select` and then releases the resource meets the trap.

## What we would expect

A drop of a `pollable` obtained from `step` releases the `pollable` only, and leaves the cache entry
in the table, which is what the platform does.

WhiteAbeLincoln:
Under the component ABI, if cache::Entry::step handed out a pollable
and the pollable was later dropped before the entry instance was,
subsequent use of the entry resulted in a panic because the async item
had been dropped from the sandbox.

The compute service doesn't have this issue because it checks the kind
of associated AsyncItem in the sandbox when dropping a pollable. Some
kinds are owned by the pollable handle, and some are owned by another
resource, with the pollable acting as a weak handle.

Porting this fix directly from the compute service would result in
breaking changes in viceroy-lib's public API (though I don't know if
this API is intended to be public or it's accidental). Additionally,
adding more items to the `AsyncItem` enum would model types which aren't
yet supported by Viceroy. Instead, a boolean flag is added to the
`PendingCacheTask` struct to indicate whether this item is shared (and
so can't be dropped through a pollable handle) or owned solely by the
pollable.

Closes #696


WhiteAbeLincoln:
@harmony7 @cceckman-at-fastly I'd appreciate a review when you get time - no rush. I suggest you review commit by commit though since there's a bulk format of the test fixtures in there which will add a bunch of noise.

This PR adds support for wasm32-wasip2 test fixtures (since afaik the issue doesn't occur under an adapted wasip1 component). That's loosely based off of wasmtime's [test-programs](https://github.com/bytecodealliance/wasmtime/tree/main/crates/test-programs) crate with the addition of a manifest to control the architecture (and maybe other build flags in the future? One of the tests was failing when built under opt = 1). This is more complex than just listing the fixture separately in the Makefile, but I preferred it since it ensures the test fixtures get automatically built when they've changed and you re-run a test, and it means a `cargo build -p test-fixtures-artifacts` works directly. I can simplify or split into a separate PR if this isn't acceptable.

The fix itself is in https://github.com/fastly/Viceroy/pull/706/changes/d6ef4edf76bb0ddf7c1b8aab4263f510c3243847. This is similar to the equivalent code in Compute & Charles's [PR 600](https://github.com/fastly/Viceroy/pull/600), though I was trying to avoid unnecessary breaking changes and so didn't add new variants to the AsyncItem enum.

.zed/settings.json
```diff
@@ -1,14 +1 @@
...
-          "Cargo.toml",
-          "test-fixtures/Cargo.toml",
-          "wasm_abi/adapter/Cargo.toml",
-          "cli/tests/trap-test/Cargo.toml"
-        ]
-      }
-    }
-  }
-}
+{}
```
cceckman-at-fastly: This seems like something that shouldn't be checked in in the first place?

